PR #23, opened 2020-07-30
Feat(engine): update to bcrypt for windows. 

tensor-programming:
# Description of change

Please write a summary of your changes and why you made them.

Updates the windows random API to bcrypt from wincrypt which is deprecated. 

## Type of change

- [ ] Documentation Fix

Ran the test suite.  

## Change checklist

### DIFF
--- a/engine/random/c_src/rng.c
+++ b/engine/random/c_src/rng.c
@@ -11,13 +11,12 @@
 #include <Security/SecRandom.h>
 #elif defined(USE_CRYPTGENRANDOM)
 #include <windows.h>
-#include <Wincrypt.h>
+#include <Bcrypt.h>
 #elif defined(USE_DEV_RANDOM)
 #include <stdio.h>
 #endif
 
-// Get a secure series of random bytes and pass the length into a buffer.
-// returns 0 upon success and 1 upon error.
+// Get a secure series of random bytes and pass the length into a buffer.returns 0 upon success and 1 upon error.
 uint8_t os_random_secrandom(uint8_t *buf, size_t len)
 {
 #if defined(USE_GETRANDOM)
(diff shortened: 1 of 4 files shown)


engine/random/c_src/rng.c
```diff
@@ -28,13 +27,21 @@ uint8_t os_random_secrandom(uint8_t *buf, size_t len)
 #elif defined(USE_SECRANDOM)
     return SecRandomCopyBytes(kSecRandomDefault, len, buf) == errSecSuccess ? 0 : 1;
 #elif defined(USE_CRYPTGENRANDOM)
-    HCRYPTPROV rng;
-    if (CryptAcquireContext(&rng, NULL, NULL, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT | CRYPT_SILENT) == 0)
+    BCRYPT_ALG_HANDLE algo_handle;
+    NTSTATUS ret1 = BCryptOpenAlgorithmProvider(&algo_handle, BCRYPT_RNG_ALGORITHM,
```
rajivshah3, line 31: Is it necessary to initialize the provider if `algo_handle` is not being passed to `BCryptGenRandom`?

tensor-programming: Did I forget to pass it?

tensor-programming:
@nothingismagick the formatting is all from stronghold not from parti.

engine/random/c_src/rng.c
rajivshah3: > Did I forget to pass it?

Nope, I think it's just not needed here. The [docs](https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom) say the `hAlgorithm` parameter should be `NULL` when using `BCRYPT_USE_SYSTEM_PREFERRED_RNG` so it seems like you don't need a handle at all

tensor-programming: Figured that one out after I wrote it.  Had it removed in an earlier version but added it back in by accident but forgot to actually put it in as a argument.

PR #31, opened 2020-08-29
feat(stronghold) add constructor to set snapshot path

lucasfernog:
# Description of change

Adds a constructor to the `Stronghold` struct so it's possible to set the snapshot path. This is crucial for systems such as Android, where the current `$HOME/.engine/snapshots` path doesn't exist.

## Type of change

Wasn't tested since we don't have unit tests yet, but I've made the same change on the wallet lib and it works there.

## Change checklist

### DIFF
--- a/stronghold/src/lib.rs
+++ b/stronghold/src/lib.rs
@@ -19,6 +19,7 @@ mod account;
 
 /// Stronghold Storage Module
 mod storage; // storage will be saving records with accounts as jsons
+pub use storage::Id;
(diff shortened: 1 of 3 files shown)


lucasfernog:
Do you think I should keep the current $HOME/.engine/snapshot path as a default? I don't think so since stronghold will be used in many operating systems and having that kind of default might confuse devs.

If we want to remove this default, I think we should delete the `engine/snapshot/src/files.rs` file.

PR #39, opened 2020-09-22
Revise the engine implementation

rootmos:
I'm getting closer to completing my overhaul of the innards of the engine, I'm creating this PR now to gather your feedback and comments.

# Description of change

The `DBView` is now a:
* set of transactions (`HashMap<TransactionId, Transaction>`)
* with cached aggregates of a stateless chain analyzing algorithm:
  - https://github.com/iotaledger/stronghold.rs/blob/b84dfdc2fedf2d043b3d540f31bb3be5f9ddc525/engine/vault/src/vault/chain.rs#L19-L25
  - https://github.com/iotaledger/stronghold.rs/blob/b84dfdc2fedf2d043b3d540f31bb3be5f9ddc525/engine/vault/src/vault/chain.rs#L57
* and a cached set of data blobs (`HashMap<BlobId, SealedBlob>`)

By looking at the properties the chains, I refactored the `DBReader::prepared_read` function signature:
```rust
pub enum PreparedRead {
    CacheHit(Vec<u8>),
    CacheMiss(ReadRequest),
    RecordIsEmpty,
    NoSuchRecord,
}

impl<'a, P: BoxProvider> DBReader<'a, P> {
    pub fn prepare_read(&self, record: &RecordId) -> crate::Result<PreparedRead> {
        // ...
    }
}
```

The different cases of the enum elucidate the major points:
* `CacheHit` happens when the `DBView` is loaded with blobs from the storage, which gives the user the ability to choose what (if anything) gets cached. The normal behavior would be to only load transactions from the storage: https://github.com/iotaledger/stronghold.rs/blob/b84dfdc2fedf2d043b3d540f31bb3be5f9ddc525/engine/examples/commandline/src/connection.rs#L55
* `CacheMiss` is the old code path
* `RecordIsEmpty` is more interesting: this covers the case of a chain with only an init transaction, so the record exists but contains no data
* `NoSuchRecord` of course means that the view haven't seen any transactions belonging to that record

One question/idea I have in my mind is to do away with the init transactions: that would remove the "empty records" concept (and also clear up the question: what happens when writing data to an uninitialized chain? currently those transactions are considered as garbage). Thoughts on this @tensor-programming?

## Other noteworthy changes
* I started using the transaction id to authenticate the decrypted transaction payload (the `encrypt`/`decrypt` methods' `ad` argument) (and similar the blobs id to authenticate the blobs), this means that the path from record id down to the blob can be trusted (the last part here is still a TODO see below)

# TODOs
* [x] maintain a `HashMap<BlobId, Vec<TransactionId>>` in order to:
  - properly garbage collect blobs that belong to several transactions https://github.com/iotaledger/stronghold.rs/blob/b84dfdc2fedf2d043b3d540f31bb3be5f9ddc525/engine/vault/src/vault.rs#L181
  -  ensure the read blob belongs to a valid transaction https://github.com/iotaledger/stronghold.rs/blob/b84dfdc2fedf2d043b3d540f31bb3be5f9ddc525/engine/vault/src/vault.rs#L236
* [x] write a set of unit tests for the now more self-contained chain pruning/analyzing algorithm, these will serve as the explanation/documentation/specification of how we want the transactions to behave
* [ ] post initial feedback, apply changes to the rest of the codebase
* [ ] add methods to return write results back to the view in order to apply updates without rereading the whole storage


tensor-programming:
Looks good, I like the direction you are headed in here.

tensor-programming (commented):
One thing Ill point out, is to be careful of `Arc` and `RC` because serialization can be expensive in serde. You can find more information here: https://serde.rs/feature-flags.html.  Basically it has to clone the data multiple times in the output which can make it much larger than anticipated. 

Everything looks good thus far.

nothingismagick:
> One question/idea I have in my mind is to do away with the init transactions: that would remove the "empty records" concept (and also clear up the question: what happens when writing data to an uninitialized chain? currently those transactions are considered as garbage).

What I think is useful about the "empty record" is that once we start 'borrowing' between strongholds, one stronghold could ask another stronghold to prove that the record is empty (if the borrow is revoked, for example)...

rootmos:
> What I think is useful about the "empty record" is that once we start 'borrowing' between strongholds, one stronghold could ask another stronghold to prove that the record is empty (if the borrow is revoked, for example)...

That's a very good use-case, "empty records" and init transactions stay!

tensor-programming:
> > What I think is useful about the "empty record" is that once we start 'borrowing' between strongholds, one stronghold could ask another stronghold to prove that the record is empty (if the borrow is revoked, for example)...
> 
> That's a very good use-case, "empty records" and init transactions stay!

Sorry missed the question for some reason.  

That was actually the original intention.  There needs to be some kind of symbolic representation for an empty chain in the case where it has been revoked or in the case where a user wants to borrow a different one (if the target chain already has data in it, it would overwrite that data if the system just copied the remote data into the stronghold).  Instead, the remote chain could verify that the target chain is empty by looking at the empty transaction.  

One other thing to note, empty transactions are also useful for obfuscation.  If you have real data in your stronghold, its harder to get to it if there are a few empty transactions also interspersed throughout the chain.

rootmos:
@mrbianchi @tensor-programming @nothingismagick I'm getting to the point that I would like to merge this, there's still work that needs to be done (with regards to the garbage collection, tests and the removal of blobs). The merge is still targeting the `feat/refactoring` branch so we can finalize the combined result there. Please note the new set of [functional unit tests of the vault](https://github.com/iotaledger/stronghold.rs/pull/39/files#diff-edbc6e54c8d8fdf7e4f010f6faae7ae1). I would like that we put a similar set of tests wrapping the stronghold (high-level implementation) @mrbianchi.

nothingismagick:
Feel free to merge @rootmos

nothingismagick (approved):

PR #42, opened 2020-10-13
Secure memory allocator and restricted runtime

rootmos:
> "Security in software sucks, but let's suck less"

The aim of this PR is to provide a unobtrusive abstraction to [guarded and locked memory allocations](https://doc.libsodium.org/memory_management#guarded-heap-allocations) and syscall filtering ([seccomp](https://en.wikipedia.org/wiki/Seccomp) on Linux).

```rust
let st: Secure<T> = Secure::new(...);

let b = zone::soft(|| {
    // We're now in an isolated process running in a restricted zone.
    //
    // No syscalls will return except:
    // - exit and write(1) (for propagating the result to the parent)
    // - anonymous mmap (and others to support the secure memory allocator)
    // - other uncontroversial auxillary syscalls (like clock_gettime and such)

    st.unlock(|t| {
         let a = t.decrypt();
         sensitive_operation(a)
    })
});
```

For the uninitiated: a syscall filtering makes it possible for a process to give up its rights to access operating system resources. As an easy example: why should a crypto algorithm ever need to access the file system (remember in Unix everything is a file)? The wrapper I implement here actually forks away from the main process (think the wallet application/stronghold daemon, that may contain UI and/or networking code and such) to perform the sensitive operations. This also adds the "normal" process isolation benefits with regards to memory.


tensor-programming:
`I, robot, ran rustfmt` Perfect!!

nothingismagick (approved):
There are a few platform problems (windows), but it needs to be merged so the client can use it.

PR #47, opened 2020-10-19
Feat p2p

elenaf9:
# Description of change

Enable communication between peers by using rust-libp2p and a mailbox concept for peers that can not communicate directly.

## Type of change

## Change checklist

- [x ] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas


engine/communication/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+//! for stronghold by using a mailbox concept that is described later.
+//! 
+//! ## Transport and the Swarm
+//! 
+//! Libp2p uses the term `transport` for their lowest layer that is responsible for sending and receiving data over a network.
+//! The current rust implementation supports tcp and websockets, and apart from that provides the option to upgrade a
+//! connection with protocols for multiplexing and authentication. 
+//! This stronghold-communication library uses yamux for multiplexing and the noise-protocol for authentication.
+//! 
+//! The second important concept of libp2p is it's `Swarm` (in newer implementations and documents also called `Switch`).
```
nothingismagick, line 30:
```suggestion
//! The second important concept of libp2p is its `Swarm` (in newer implementations and documents also called `Switch`).
```

engine/communication/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+//! The main entry point for all communication with other peers is the `P2PNetwork`. 
+//! It creates the transport and the swarm for the prior created `P2PNetworkBehaviour` and listens for incoming connections.
+//! It has multiple listening addresses due to libp2ps concept of `multiaddresses` that encode different addressing schemes for different
+//! protocols. Apart from IPv4 and IPv6 Addresses, these multiaddresses can also be dns addresses, which is relevant if a peer is listening
+//! to such an address on a server. The listed multiaddresses are only the ones within the same local network, but if port forwarding was configured,
+//! the local /ip4/my-local-address/tcp/12345 Address can be replaced by the public one or by `/dns/my.public.server.address/tcp/12345`, where the 
+//! `/tcp/12345` part describes the port.
+//! 
+//! 
+//! It provides methods for dialing and connection other peers, using the swarm behaviour via it's `swarm` property (that enables using methods
```
nothingismagick, line 55:
```suggestion
//! It provides methods for dialing and connection other peers, using the swarm behaviour via its `swarm` property (that enables using methods
```

nothingismagick (changes requested):
We need to discuss how we want to propagate connection errors as well as removing the println! macro usage because we cannot assume that all library consumers will have a console at all.

engine/communication/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# build mailbox
+RUN cargo build --example mailbox
+FROM rust:latest
+WORKDIR /app
+
+# copy the build artifact from the build stage
+COPY --from=build /stronghold-communication/target/debug/examples/mailbox .
+ENTRYPOINT ["/app/mailbox"]
+CMD ["start-mailbox"]
```
nothingismagick, line 35:
```suggestion
CMD ["start-mailbox"]

```

engine/communication/src/behaviour/protocol.rs
```diff
@@ -0,0 +1,206 @@
...
+// TODO: support no_std
+use std::io::{Cursor as IOCursor, Error as IOError, ErrorKind as IOErrorKind, Result as IOResult};
+
+/// Custom protocol that extends libp2p's RequestReponseProtocol
+#[derive(Debug, Clone)]
+pub struct MessageProtocol();
+
+impl ProtocolName for MessageProtocol {
+    fn protocol_name(&self) -> &[u8] {
+        b"/p2p-mailbox/1.0.0"
```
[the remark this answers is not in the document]
elenaf9: How can you do this?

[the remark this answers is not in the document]
nothingismagick: Read in from a file at build time?

engine/communication/src/message.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MailboxRecord {
+    key: String,
+    value: String,
+    expires_sec: u64,
+}
+
+impl MailboxRecord {
+    pub fn new(key: String, value: String, expires_sec: u64) -> Self {
```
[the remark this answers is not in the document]
nothingismagick: quoting Lucas:

> Exposing `u64` values to JavaScript will result in the use of [BigUint64Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigUint64Array) in the generated JS wrapper code, and it isn't supported on Safari. So it's recommended to use `f64` instead, see [this PR on iota.rs](https://github.com/iotaledger/iota.rs/pull/94).

engine/communication/src/behaviour/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+    ///       _event: RequestResponseEvent<Request, Response>,
+    ///    ) {}
+    ///
+    ///    fn handle_kademlia_event(_swarm: &mut impl SwarmContext, _result: KademliaEvent) {}
+    /// }
+    ///
+    /// let behaviour = P2PNetworkBehaviour::<Handler>::new(local_keys.public()).unwrap();
+    /// ```
+    pub fn new(_public_key: PublicKey) -> QueryResult<Self> {
+        let peer_id = PeerId::from(_public_key);
```
tensor-programming, line 203: should remove the underscore.  Only really exists if you don't plan to use the argument which you are using here.

engine/communication/src/behaviour/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+            inner: PhantomData,
+        })
+    }
+}
+
+impl<C: InboundEventCodec + Send + 'static> NetworkBehaviourEventProcess<MdnsEvent> for P2PNetworkBehaviour<C> {
+    // Called when `mdns` produces an event.
+    fn inject_event(&mut self, _event: MdnsEvent) {
+        #[cfg(feature = "kademlia")]
+        if let MdnsEvent::Discovered(list) = _event {
```
tensor-programming, line 233: Again remove the underscore from event.

engine/communication/src/network/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                .ok_or_else(|| QueryError::Mailbox(format!("No know mailbox for {}", peer_id)))
+        } else {
+            Ok(mailboxes.get_default())
+        }?;
+        Ok(self.swarm.send_request(peer, Request::Publish(record)))
+    }
+}
+
+#[cfg(test)]
+struct DummyHandler();
```
tensor-programming, line 197: can remove the parenthesis.  Doesn't matter either way.

tensor-programming (changes requested):

engine/communication/src/message.rs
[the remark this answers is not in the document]
elenaf9: since the u64 is required by kademlia and for now this wont be exposed to the outside of the application, I'd ignore this for now but keep it in mind for later changes

engine/communication/src/behaviour/mod.rs
```diff
@@ -85,6 +88,7 @@ impl P2PNetworkBehaviour {
         let mdns =
             Mdns::new().map_err(|_| QueryError::ConnectionError("Could not build mdns behaviour".to_string()))?;
 
+        let identify = Identify::new("/ipfs/0.1.0".into(), "rust-ipfs-example".into(), local_keys.public());
```
nothingismagick, line 91: ipfs?

nothingismagick (approved):

engine/communication/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+bytes = "0.5.6"
+clap = { version = "3.0.0-beta.1", features = ["yaml"] }
+futures = "0.3.1"
+libp2p = {version = "0.28.1", default-features = false, features = ["dns", "identify", "mdns-async-std", "mplex", "noise", "request-response", "tcp-async-std", "yamux", "websocket"]}
+prost = {version = "0.6.1", default-features = false, features = ["prost-derive"] }
+regex = "1.3.9"
+thiserror = "1.0.21"
+serde = { version = "1.0.117", default-features = false, features = ["alloc", "derive"] }
+serde_json = { version = "1.0.59", default-features = false, features = ["alloc"] }
+riker = "0.4.1"
```
nothingismagick, line 21: Riker 0.4.2 was just published, so this is now out of date. In order to stay 'synced' with the client deps, please drop the patch versions for now. At some point we may return to this, but I am concerned that things stop working in concert if we don't have parity.

engine/communication/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "communication"
```
nothingismagick, line 2:
```suggestion
name = "stronghold-communication"
```

engine/communication/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version = "0.1.0"
+authors = ["Elena Frank <elena.frank@iota.org>"]
+edition = "2018"
+license = "Apache-2.0 OR MIT"
```
nothingismagick, line 6:
```suggestion
license = "Apache-2.0"
```

communication/examples/local-echo.rs
```diff
@@ -0,0 +1,242 @@
...
+                }
+            }
+        }
+        Poll::Pending
+    }));
+    Ok(())
+}
+
+fn handle_input_line(swarm: &mut Swarm<P2PNetworkBehaviour<Request, Response>>, line: String) {
+    if let Some(peer_id) = Regex::new("PING\\s+\"(\\w+)\"")
```
tensor-programming, line 200: Little verbose on the if branches.  Would be better if you put most of the internal logic into the body of the if checks so that it doesn't have to execute it every time.

tensor-programming (changes requested):

tensor-programming (approved):

PR #51, opened 2020-11-16
Feat/add client

tensor-programming:
# Description of change

High level client library

## Type of change

- [ ] Documentation Fix

## Change checklist


nothingismagick:
A couple of comments regarding logging / debug:

I think we should add a `/config/riker.toml` so that we can have some more granular control over how much we log. See: https://riker.rs/config/

Also, we should definitely use https://docs.rs/bee-common-derive/0.1.0-alpha/bee_common_derive/derive.SecretDebug.html

`#[derive(SecretDebug)]` instead of normal debug derives...

tensor-programming:
> A couple of comments regarding logging / debug:
> 
> I think we should add a `/config/riker.toml` so that we can have some more granular control over how much we log. See: https://riker.rs/config/
> 
> Also, we should definitely use https://docs.rs/bee-common-derive/0.1.0-alpha/bee_common_derive/derive.SecretDebug.html
> 
> `#[derive(SecretDebug)]` instead of normal debug derives...

Will look into all of these things after I get the PR up and running properly.

nothingismagick (changes requested):
just a couple cosmetic things

client/src/actors.rs
```diff
@@ -0,0 +1,267 @@
...
+#[derive(Debug, Clone)]
+pub enum BMsg<P: BoxProvider + Debug> {
+    CreateVault(VaultId, Key<P>),
+    ReadData(Key<P>, RecordId),
+    WriteData(Key<P>, RecordId, Vec<u8>, RecordHint),
+    InitRecord(Key<P>, VaultId),
+    RevokeData(Key<P>, RecordId),
+    GarbageCollect(Key<P>),
+    ListAsk(Key<P>),
+    WriteSnapshot(String, Option<String>, Option<PathBuf>),
```
lucasfernog, line 30: I understand what most of the enum values means, but in some cases maybe named arguments would be better? I have no idea what is the Option<String> here without looking at the rest of the code.

client/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[error("Vault Error: {0}")]
+    VaultError(#[from] engine::vault::Error),
+}
+
+/// Creates the ActorSystem for stronghold and attaches the actors.  Returns the ActorSystem and the Channel.
+pub fn init_stronghold() -> (ActorSystem, ChannelRef<SHResults>) {
+    let sys = ActorSystem::new().unwrap();
+    let chan: ChannelRef<SHResults> = channel("external", &sys).unwrap();
+
+    sys.actor_of_args::<Client, _>("client", chan.clone()).unwrap();
```
lucasfernog, line 62: I think we should prefix the actor names with `stronghold-internal-` just so we won't face weird issues if consumers add their own actors and one of them is named "client".

note that changing this also means that we need to change the `ctx.select` calls.

client/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(DeriveError, Debug)]
+pub enum Error {
+    #[error("Id Error")]
+    IDError,
+    #[error("Vault Error: {0}")]
+    VaultError(#[from] engine::vault::Error),
+}
+
+/// Creates the ActorSystem for stronghold and attaches the actors.  Returns the ActorSystem and the Channel.
+pub fn init_stronghold() -> (ActorSystem, ChannelRef<SHResults>) {
```
lucasfernog, line 58: for flexibility it might be a good idea to expose a function that takes the ActorSystem and injects all stronghold actors in there.

client/src/actors.rs
```diff
@@ -0,0 +1,267 @@
...
+        self.receive(ctx, msg, sender);
+    }
+}
+
+impl Receive<SMsg> for Snapshot {
+    type Msg = SMsg;
+
+    fn receive(&mut self, ctx: &Context<Self::Msg>, msg: Self::Msg, _sender: Sender) {
+        match msg {
+            SMsg::WriteSnapshot(pass, name, path, state) => {
```
lucasfernog, line 105: does this mean I need to manually tell stronghold when I want to save the pending changes to a snapshot file?

client/src/actors.rs
tensor-programming: Will put documentation above each of the variant types.  For instance, in this one, `WriteSnapshot` takes the password, an optional Filename, and the Optional File Path.

client/src/actors.rs
tensor-programming: For now, there is no periodic "write" job.  However, that is something on the priority list.  To add functionality to allow you to specify that duration.  I also may add some features to allow the writes to happen after every single change to the state of the snapshot.

client/src/lib.rs
tensor-programming: Good idea.

nothingismagick:
hello robot

nothingismagick (approved):

tensor-programming:
can merge whenever.

rootmos:
Here's the fix for the broken test: https://github.com/iotaledger/stronghold.rs/pull/59

PR #70, opened 2020-12-08
Rework the snapshot format and its implementation

rootmos:
Change the snapshot format to:
* remove responsibility for hashing the user password, in favor of a small side library with suggested parameters
* expose the `assocated_data` to potentially be used as another layer of protection, perhaps use something end-user unique, or perhaps application specific
* remove libsodium (sodiumoxide) dependency in favor of crypto.rs: it's crazy slow to build and the main functionality we require is already provided by crypto.rs
* non-streaming encryption
  - for a simpler initial snapshot format and implementation
  - benchmark performance in situations when encryption rivals the non-`sync`:ed `write` syscall, here my reasoning is: when data isn't forced to be written to disk, the write syscall would only move the data into it's own memory backed write cache, what snapshot sizes warrant a streaming implementation? Consider here that I consider the streaming version of xchachapoly1305 from libsodium (secretstreams) to be as secure in an offline setting since an attacker would have access to the "whole stream".

Implement the "write to temporary file and swap" trick:
- this should alleviate the file corruption issues we've observed
- can this be done so that non-file system files can also be used? i.e. a raw interface with `fd`:s and a high-level interface that does the swap-trick?


engine/snapshot/src/logic.rs
```diff
@@ -1,304 +1,112 @@
...
+use crypto::{
+    ciphers::chacha::xchacha20poly1305,
+    rand,
+};
 
-/// create a decryption pull stream.
-fn pull_stream(header: &[u8], &Key(ref key): &Key) -> crate::Result<Stream<Pull>> {
-    let stream_key = secretstream::Key(key.to_owned());
-    let header = Header::from_slice(header).expect("Invalid Header size");
+/// PARTI in binary
```
nothingismagick, line 14: ah, nostalgia

nothingismagick (approved):

PR #88, opened 2020-12-17
chore(communication): bump libp2p version

elenaf9:
# Description of change

Patch libp2p version from `v0.28` to `v0.32`.

fixes issue #73 

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/communication/src/actor.rs
+++ b/communication/src/actor.rs
@@ -215,7 +215,7 @@ fn poll_swarm<T: MessageEvent, U: MessageEvent>(
             }
             CommunicationEvent::GetSwarmInfo => {
                 if let Some(sender) = sender_opt {
-                    let peer_id = Swarm::local_peer_id(&swarm).clone();
+                    let peer_id = *Swarm::local_peer_id(&swarm);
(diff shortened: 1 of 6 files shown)


communication/src/actor.rs
```diff
@@ -377,8 +377,9 @@ mod test {
             {
                 self.has_received_response = true;
             } else if let CommunicationEvent::ConnectPeerResult { addr: _, result } = msg {
+                let peer_id = result.expect("Panic due to no network connection");
```
nothingismagick, line 380: Nice one. I assume this resolves the issue I faced with little-snitch blocking...

nothingismagick (approved):

nothingismagick:
Can you add a .changes file @elenaf9 ?

communication/src/actor.rs
elenaf9: The thing is that this test can only be successful if the process can use localhost, so if little-snitch is preventing that, it is correct that this should panic. But at least there is now a proper error message on that, if we want to handle this in a better way, I will have to either add a flag or some sort of delay to give little-snitch the time to allow it.

PR #122, opened 2021-01-08
Add back the atomic Ed25519 procedures

rootmos:
@lucasfernog:
I've renamed the previously incorrectly named `Ed25519PublicKey` and `Ed25519Sign` procedures:
- `Ed25519PublicKey` -> `SLIP10DeriveAndEd25519PublicKey`
- `Ed25519Sign` -> `SLIP10DeriveAndEd25519Sign`
in order to give room for adding back the atomic Ed25519 procedures.

The previous (combined) procedures use bee-p`s ed25519 signer and so enforces 256-bit seeds, something that we do not want to do (and can't do if BIP39 is desired to be used since it uses 512 bit seeds). These new "atomic" procedures use crypto.rs's ed25519 implementation and the (currently internal) SLIP10 implementation (which will be moved to crypto.rs shortly). I've added explicit tests showing that these implementations are equivalent (and how they are intended to be used "equivalently").

I suggest we let @lucasfernog adapt to this somewhat backwards compatible change (only a rename to keep the old functionality) before we remove the combined methods (and the now redundant `SignUnlockBlock` procedure, which is doing the same as `SLIP10DeriveAndEd25519Sign` procedure).

Another noteworthy change is that SLIP10Generate now requires an explicit size of the generated seed (I suggest we make this optional and default to 512 bits), this was done in order to make above mentioned equivalence tests.

Internally I've migrated the random test data generation (poor-man's quickcheck) tools to a top-level `test_utils` crate, the naming and placement of this crate I leave open to discussion and further improvement.


nothingismagick (changes requested):
missing a .changes file

nothingismagick:
Well, one of the things that I love about this project is that the code is SO readable that it basically negates the requirement of comments. Nevertheless, we need to start thinking about proper documentation and specification. Its not an issue for this PR, but something we need to take a dedicated look into after the firefly audit.

nothingismagick (approved):

tensor-programming (approved):

PR #123, opened 2021-01-12
Use X25519 in the snapshot format

rootmos:
Uses https://github.com/iotaledger/crypto.rs/pull/45.

For the curious: adding a key exchange with an ephemeral key-pair in front of a symmetric cipher algorithm is done in order to protect the original key in the scenario when the cipher is broken in such a way as to reveal the key (especially when having access to multiple ciphertexts makes such an attack (more) feasible). Note that this is very similar to what a nonce is in other ciphers. In our case the cipher is used to protect the offline storage, hence this is an offline attack mitigation, and so especially important since we can't guarantee that an offline attacker keeps access to old snapshot files (that the user might think have been removed).


tensor-programming (dismissed):
Make sure that clippy passes before you commit. Everything else looks good.

nothingismagick:
Can you fix the conflict and clippy @rootmos ?

nothingismagick:
Also should add a changefile

rootmos:
Rebased. Changelog on the "stronghold-engine" crate (I assume we don't publish+track its child crates individually).

nothingismagick:
@rootmos - individual subcrates may have their own changelogs, fyi. Compare:
https://github.com/iotaledger/stronghold.rs/blob/dev/engine/vault/CHANGELOG.md

lucasfernog (approved):

PR #124, opened 2021-01-12
Secrets marshaling

rootmos:
Add types necessary for conveniently protecting secret values and only open them in the secured runtime zone.

New types that have been introduced:
* in `runtime::mem::guarded`: a `GuardedBox`, `GuardedVec` and `GuardedString` that use the guarded memory allocator to lock the memory (accessing it would yield a SIGSEGV) when not in use
* in `engine::secret`: `X25519XSalsa20Poly1305` and `AES` two "boxes" that can be used to marshal secrets around in an insecure environment (they use the guarded containers to protect both keys and the unlocked data).

Example from [vault.rs](engine/vault/src/vault.rs):
```rust
let ct = ZoneSpec::default().secure_memory().random().run(|| {
    let pt = recipient_key.as_ref().access().access(data.as_ref()).unwrap();
    let ct = (&*pt.access()).encrypt(&self.view.key, blob_id).unwrap();
    ct
})?;
```


tensor-programming (changes requested):
Make sure that the runtime builds on windows even if you have to cfg out the current logic for now.

rootmos:
So, discovered last week that we need guarded versions of `Vec` and `String`, since that's how the secrets are typed when being transported around in stronghold: since it's only when the secrets are used that they get typed as `[u8; 32]` and such and can be protected by a `GuardedBox`. This change also provoked another pass at refactoring the `Protection` and `Access` traits to be more usable. Sorry for the mess, is there a way to convert this PR back into a draft until I'm certain the types will stay?

tensor-programming:
Converted it back to draft for you; its up at the top next to reviewers.  Unfortunately I also took this and rebased my PR off of it, but Ill go ahead and see if I can revert some of that.

rootmos:
Status update: the full flow is now working on Linux (and only minor technical issues on MacOS regarding EOF detection), so we can now review the types: how they feel and operate.

The use-case I implemented first is the communication between the internal actor and the vault. Next one I want to target is the BIP39's mnemonic sentence and passphrase: this would be a breaking change for the wallet @lucasfernog. You would have to first ask the stronghold for it's recipient public key and use it to `protect` the sentence and passphrase before putting it in a request, [here's how that looks in the internal actor](https://github.com/iotaledger/stronghold.rs/blob/a5bb8c1884ef5b4b94ec0b7c6933a948905e1a93/client/src/actors/internal.rs#L203).

The windows implementation should not be far behind: the guarded containers (where the compilation fails right now) only needs a corresponding guarded allocator (@tensor-programming any updates on this?).

tensor-programming:
@rootmos my allocator is mostly done, Im going to try to merge it in here and we will see if it can make the windows side work as well as the linux side.   Just going to put it on this PR for now.

rootmos:
So, managed to figure out the issue on macos:

When the value is already transferred and we only need to wait for the child process to terminate I used to use `poll` with the same (now empty/EOF:ed) file descriptor, reasoning that since there shouldn't come any more data on it `poll` essentially is just a timeout (no more `POLLIN`s), but macos seems to retrigger the `POLLHUP` event in that scenario (turning the loop into a spin-lock?). The solution was to use `nanosleep` in this scenario to perform the waiting. I tried to remove the `WNOHANG` from the `waitpid` call, but that breaks the timeout functionality, see the added tests. I guess `nanosleep` is also communicates the intent in a better way, so win-win?

tensor-programming:
closed for now in favor of #155

PR #130, opened 2021-01-16
Feat/integrate comms

### LINKED ISSUE #111: communications integration
## Transparency
The communications crate functions in and of itself, however we need to integrate it with the client. Furthermore, we need to provide guidance for implementors wishing to use the "Internet of Strongholds". 

This will require:
1. riker interface in client
2. example implementation
3. verification on USB Armory

elenaf9:
# Description of change

Refactor communication actor and integrate it into the stronghold interface.

## Type of change

Added interface tests to read/ write a remote stronghold store.

## Change checklist


nothingismagick:
@lucasfernog and @tensor-programming - would love to have you both do an in depth review of this PR, from the perspective of your respective sides of this actor stack.

elenaf9:
I am right now doing some refactoring, but I will soon be done with that and then also mark the PR's as ready for review.

lucasfernog:
Please tag me when this is ready :)

elenaf9:
@lucasfernog @tensor-programming the PR is ready for review, but the following aspects are not implemented yet / need to be discussed afterwards:
-  what methods should exists in the stronghold interface (currently I only added `write_remote_vault`, `read_remote_store` and `write_remote_store`)
- the firewall actor is at the moment just a dummy actor that accepts all requests, eventually it should be provided by the user
- my changes depends on [this](https://github.com/riker-rs/riker/pull/162) PR to riker, right now I've set my fork of riker as dependency for development purposes.

client/src/actors/client.rs
```diff
@@ -426,6 +499,11 @@ impl Receive<SHRequest> for Client {
                 let payload = self.read_from_store(vid.into());
 
                 if let Some(payload) = payload {
+                    println!(
```
lucasfernog: do we want to keep this?

client/src/interface.rs
```diff
@@ -53,11 +85,23 @@ impl Stronghold {
...
+                .actor_of_args::<CommunicationActor<_, SHResults, _, _>, _>("communication", (local_keys, config))
+                .expect(line_error!())
+        };
         system
             .actor_of_args::<InternalActor<Provider>, _>(&format!("internal-{}", id_str), client_id)
             .expect(line_error!());
 
         system.actor_of::<Snapshot>("snapshot").expect(line_error!());
+        println!("\n\n");
+        system.print_tree();
```
lucasfernog: maybe remove this one too.. or leave it under `cfg!(debug_assertions)`

client/src/actors/client.rs
elenaf9: no sorry, I only inserted that for debugging purposes, I'll remove it

client/src/interface.rs
```diff
@@ -98,16 +148,29 @@ impl Stronghold {
...
 
         if self.client_ids.contains(&client_id) {
             let idx = self.client_ids.iter().position(|cid| cid == &client_id);
 
             if let Some(idx) = idx {
                 self.current_target = idx;
-            }
 
+                #[cfg(feature = "communication")]
+                match ask(
```
lucasfernog: does this mean that every `switch_actor_target` call changes both the local stronghold client and the connected strongholds?

client/src/interface.rs
```diff
@@ -569,4 +632,231 @@ impl Stronghold {
...
+            CommunicationResults::ConnectPeerResult(Err(reason)) => {
+                return StatusMessage::Error(format!("Error connecting peer: {:?}", reason))
+            }
+            _ => return StatusMessage::Error("Invalid communication event".into()),
+        };
+
+        let vault_path = &location.vault_path();
+        let vault_path = vault_path.to_vec();
+        // check if vault exists
+        let b = match self
```
lucasfernog: can you rename these `b` variables to `vault_exists` and `record_exists`?

client/src/interface.rs
```diff
@@ -569,4 +632,231 @@ impl Stronghold {
...
+    pub async fn write_to_remote_store(
+        &self,
+        peer_id: PeerId,
+        addr: Multiaddr,
+        location: Location,
+        payload: Vec<u8>,
+        lifetime: Option<Duration>,
+    ) -> StatusMessage {
+        // connect remote peer
+        let peer_id = match ask::<_, _, CommunicationResults<SHResults>, _>(
```
lucasfernog: this `ConnectPeer` block repeats a lot, maybe it should be a separate function too ;)

client/src/interface.rs
lucasfernog: also, should this whole logic be a core stronghold message? like a `EnsureVaultExists` and `EnsureRecordExists` - maybe this is useful for consumers too

communication/src/actor/message.rs
```diff
@@ -0,0 +1,212 @@
...
+    UnbanPeer(PeerId),
+    /// Start listening to a port on the swarm. If no `Multiaddr` is provided, the address will be OS assigned.ActorRef
+    StartListening(Option<Multiaddr>),
+    /// Stop listening to the swarm. Without a listener, the local peer can not be dialed from remote.
+    RemoveListener,
+}
+
+/// The firewall that rejected or dropped the request
+#[derive(Debug, Clone)]
+pub enum FirewallRejected {
```
lucasfernog: `FirewallRejected` sounds like an error type, what about `RejectionFirewall` or something like that? or just `FirewallType`

client/src/interface.rs
tensor-programming: agreed, might be useful to make a connect_peer function.

communication/src/actor/message.rs
tensor-programming: `FirewallBlocked` perhaps?

client/src/interface.rs
elenaf9: It changes the local stronghold client and at the same time sets this new stronghold client to also be the one that the communication actor(/ Swarm task) talks to (i.g. forwards incoming request to). I could implement this in a separate method, but imo it makes sense that if the user switches the client, the remote request should be forwarded to that new client. But maybe we have to discuss this again for other use cases

communication/src/actor/message.rs
elenaf9: How about `RejectionFirewallSource` or just `FirewallSource`? 
`FirewallType` sounds like there are different types of firewalls, but actually they differ on their source/ origin (i.g. is it the local firewall or the remote one), rather than on their type. But then again maybe "type" can be interpreted in different ways.

client/src/interface.rs
elenaf9: I copied this logical sequence of the messages from the interface for targeting the local client, so this question also concerns those methods. But yes I can imagine that a user might need these dedicated message types when they don't want to automatically create a new vault if the vault path doesn't exist, especially with remote strongholds but also in a local system.
@tensor-programming what do you think about this?

tensor-programming (changes requested):
Most of this looks good.

tensor-programming (approved):

client/src/interface.rs
tensor-programming: I have to wonder if we should even allow a user to create a remote vault.  Or if there should be some kind of policy for restricting that kind of behavior.   Ideally, we would want each user to have access to a select number of vaults based on which paths they know and haven't really considered how we would restrict users inside of a shared stronghold.  Perhaps, its something to think about later. 

For now it should be ok to just enable the ability to add/create a new vault remotely if the vault doesn't exist and we can restrict/change that behavior later using policies.

client/src/interface.rs
tensor-programming: This should be fine as is. We can work with other use-cases when as they come.

client/src/interface.rs
elenaf9: Well, i think this type of policy that you are talking about Tensor could maybe be decided by the Firewall, so the Firewall can be configured for each stronghold individually to allow or reject a `SHRequest::CreateNewVault`?

elenaf9:
The tests currently fail until this [PR](https://github.com/libp2p/rust-libp2p/pull/1899) is merged.

lucasfernog (approved):

client/Cargo.toml
```diff
@@ -22,7 +22,8 @@ zeroize_derive = "1.0"
 anyhow = "1.0"
 thiserror = "1.0"
 futures = "0.3"
-riker = "0.4"
+riker = { version = "0.4", git ="https://github.com/elenaf9/riker", branch = "master" }
```
tensor-programming, line 25: Curious as to why you are pointing it to your own repo.  Is there some unmerged changed that needs to be here for the client and for the comms layer?

communication/src/actor/ask.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020-2021 IOTA Stiftung
```
tensor-programming: Maybe its worth consolidating these files into a library called `utils` or `stronghold_utils` since this code appears in the interface as well?

tensor-programming (commented):
Looks good.  Just a couple of questions and minor things really.   

One more thing, there are a ton of unwraps in the code that kind of make me uneasy.  I know that its kind of hard to deal with these especially when you are implementing an interface or working in a closure. The problem is that theses unwraps will cause panics if the error happens (in most cases these are options).  In production that's not necessarily what we want to happen, in other words, some of them should be recoverable errors if possible.   I would recommend going back over each of the unwrap cases to make sure that they are absolutely necessary. If you can't remove the unwrap, it might be better to use `expect` because at least you give the user some information.

client/Cargo.toml
elenaf9:
```
where
    Req: MessageEvent,
    T: Message + From<Req>,
{
   pub client: ActorRef<T>
```
This trait bound depends on a [PR at riker](https://github.com/riker-rs/riker/pull/162) that has been approved, but somehow not merged yet(it seems like the maintainer of the repo is currently inactive). In my fork I already merged the PR, which is why is used it as dependency. 
If this is an issue, I could change the trait restriction to sth like this (which would be possible without the PR merged):
```
where
    Req: MessageEvent,
    T: Tell<Req> + Clone + Debug + ... ,
{
   pub client: T
```

communication/src/actor/ask.rs
elenaf9: We already have the `test_utils` library, maybe we could rename it to `utils` and have separate modules in it for `test` stuff and `actor` or something? Or is it better to separate it completely

client/Cargo.toml
tensor-programming: Alright, that's what I suspected.  Its fine as it is then.

communication/src/actor/ask.rs
tensor-programming: Hmm, perhaps it could be done that way.  And actually lets just go with that.  Just rename it to `utils` and put the ask.rs files in there. If you want, we can do it in a separate PR or do it now.

PR #158, opened 2021-03-04
Feat/communication firewall

elenaf9:
# Description of change

Integrate the firewall into the communication actor, instead of expecting the user to provide a firewall actor.
Add procedural macro derive to generate permission for a Request  and create unit variant of the request, e.g.:
```
pub enum Request {
    VariantA { inner : u32 },
    VariantB(String),
    VariantC
}
```
will generate 

```
impl ToPermissionVariants<RequestPermission> for Request {
    fn to_permission_variants(&self) -> RequestPermission {
        match self {
            Request::VariantA {inner: _} => RequestPermission::VariantA,
            Request::VariantB(_) => RequestPermission::VariantB,
            Request::VariantC=> RequestPermission::VariantC,
        }
    }
}

#[derive(Clone, Debug)]
pub enum RequestPermission {
    VariantA,
    VariantB,
    VariantC
}

impl VariantPermission for RequestPermission {
    fn variant_permission_value(&self) -> PermissionValue {
        let n = match self {
            VariantA => 0,
            VariantB => 1,
            VariantC => 2
        };
        PermissionValue::new(n)
    }
}
```

This allows setting different permissions for each request variant (similar to how read/write/execute permissions in linux work).
The firewall takes a sum (e.g. 5) and check with that if a request is allowed (in this case VariantA and VariantC would be, Variant B not).

## Type of change

Existing firewall tests, and new tests for the permission concept.

## Change checklist


nothingismagick (changes requested):
I really like this approach and don't have any problems with it. I think it will be important to document the default behaviour of the Firewall, which is to allow any outgoing connection, but deny any incoming.

I would also like to see a README in the `communication-macros` folder.

client/src/interface.rs
```diff
@@ -729,7 +731,26 @@ impl Stronghold {
...
+    #[cfg(feature = "communication")]
+    /// Read from the store of a remote stronghold
+    pub async fn configure_firewall(&self, rule: FirewallRule) -> StatusMessage {
+        if self.communication_actor.is_none() {
+            return StatusMessage::Error(String::from("No communication spawned"));
+        }
+
+        match ask::<_, _, CommunicationResults<SHResults>, _>(
+            &self.system,
+            self.communication_actor.as_ref().unwrap(),
```
tensor-programming, line 748: Wonder if any of these unwraps can be removed.

communication/communication-macros/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+                    #match_variants
+                }
+            }
+        }
+    }
+}
+
+fn impl_permission(name: &Ident, data_enum: &DataEnum) -> TokenStream {
+    if data_enum.variants.len() > 32 {
+        panic!("More then 32 variants on enums are not allowed.");
```
tensor-programming, line 124: I wonder if this should be here or if it should be a recoverable error.

tensor-programming (changes requested):
Looks good to go.

tensor-programming (approved):

nothingismagick (approved):

PR #210, opened 2021-05-07
Feat/comms behaviour protocol

### LINKED ISSUE #72: RUSTSEC-2020-0057: crate has been renamed to `cipher`
> crate has been renamed to `cipher`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `block-cipher`                      |
| Version             | `0.8.0`                   |
| URL                 | [https://github.com/RustCrypto/traits/pull/337](https://github.com/RustCrypto/traits/pull/337) |
| Date                | 2020-10-15                         |

This crate has been renamed from `block-cipher` to `cipher`.

The new repository location is at:

&lt;https://github.com/RustCrypto/traits/tree/master/cipher&gt;

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2020-0057.html) for additional details.

### LINKED ISSUE #74: RUSTSEC-2020-0058: crate has been renamed to `cipher`
> crate has been renamed to `cipher`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `stream-cipher`                      |
| Version             | `0.7.1`                   |
| URL                 | [https://github.com/RustCrypto/traits/pull/337](https://github.com/RustCrypto/traits/pull/337) |
| Date                | 2020-10-15                         |

This crate has been renamed from `stream-cipher` to `cipher`.

The new repository location is at:

&lt;https://github.com/RustCrypto/traits/tree/master/cipher&gt;

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2020-0058.html) for additional details.

### LINKED ISSUE #207: RUSTSEC-2021-0060: merged into the `aes` crate
> merged into the `aes` crate

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `aes-soft`                      |
| Version             | `0.5.0`                   |
| URL                 | [https://github.com/RustCrypto/block-ciphers/pull/200](https://github.com/RustCrypto/block-ciphers/pull/200) |
| Date                | 2021-04-29                         |

The `aes-soft` crate has been merged into the `aes` crate. The new repository
location is at:

&lt;https://github.com/RustCrypto/block-ciphers/tree/master/aes&gt;

AES-NI is now autodetected at runtime on `i686`/`x86-64` platforms.
If AES-NI is not present, the `aes` crate will fallback to a constant-time
portable software implementation.

To force the use of a constant-time portable implementation on these platforms,
even if AES-NI is available, use the new `force-soft` feature of the `aes`
crate to disable autodetection.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0060.html) for additional details.

### LINKED ISSUE #208: RUSTSEC-2021-0059: merged into the `aes` crate
> merged into the `aes` crate

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `aesni`                      |
| Version             | `0.8.0`                   |
| URL                 | [https://github.com/RustCrypto/block-ciphers/pull/200](https://github.com/RustCrypto/block-ciphers/pull/200) |
| Date                | 2021-04-29                         |

The `aesni` crate has been merged into the `aes` crate. The new repository
location is at:

&lt;https://github.com/RustCrypto/block-ciphers/tree/master/aes&gt;

AES-NI is now autodetected at runtime on `i686`/`x86-64` platforms.
If AES-NI is not present, the `aes` crate will fallback to a constant-time
portable software implementation.

To prevent this fallback (and have absence of AES-NI result in an illegal
instruction crash instead), continue to pass the same RUSTFLAGS which were
previously required for the `aesni` crate to compile:

```
RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
```

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0059.html) for additional details.

### LINKED ISSUE #212: Communication types
Can we please have Display on the error types and maybe std::error::Error traits too... 

If this is part of the open PR, that's great and feel free to close.

### LINKED ISSUE #228: RUSTSEC-2021-0073: Conversion from `prost_types::Timestamp` to `SystemTime` can cause an overflow and panic
> Conversion from `prost_types::Timestamp` to `SystemTime` can cause an overflow and panic

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `prost-types`                      |
| Version             | `0.7.0`                   |
| URL                 | [https://github.com/tokio-rs/prost/issues/438](https://github.com/tokio-rs/prost/issues/438) |
| Date                | 2021-07-08                         |
| Patched versions    | `>=0.8.0`                  |

Affected versions of this crate contained a bug in which untrusted input could cause an overflow and panic when converting a `Timestamp` to `SystemTime`.

It is recommended to upgrade to `prost-types` v0.8 and switch the usage of `From&lt;Timestamp&gt; for SystemTime` to `TryFrom&lt;Timestamp&gt; for SystemTime`.

See [#438] for more information.

[#438]: https://github.com/tokio-rs/prost/issues/438

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0073.html) for additional details.

### LINKED ISSUE #230: RUSTSEC-2021-0076: libsecp256k1 allows overflowing signatures
> libsecp256k1 allows overflowing signatures

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `libsecp256k1`                      |
| Version             | `0.3.5`                   |
| URL                 | [https://github.com/paritytech/libsecp256k1/pull/67](https://github.com/paritytech/libsecp256k1/pull/67) |
| Date                | 2021-07-13                         |
| Patched versions    | `>=0.5.0`                  |

libsecp256k1 accepts signatures whose R or S parameter is larger than the
secp256k1 curve order, which differs from other implementations. This could
lead to invalid signatures being verified.

The error is resolved in 0.5.0 by adding a `check_overflow` flag.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0076.html) for additional details.

elenaf9:
# Description of change

Implementation of a custom Network Behaviour protocol for stronghold-communication.

Replace the currently used Network Behaviour with a custom protocol. The new protocol follows the basic concepts of the libp2p-request-response protocol, but additionally implements a firewall and directly integrates and leverages libp2p-relay and libp2p-mdns.

## Type of change

- [x] Documentation Fix

## Change checklist


mxinden:
Great to see usage of rust-libp2p in general and libp2p-relay in specific! I would be really curious to hear any feedback on libp2p-relay (and of course on the greater rust-libp2p project). Also note the effort around [circuit relay v2 implementation](https://github.com/libp2p/rust-libp2p/pull/2059).

mxinden:
In case you are interested in hole punching for stronghold.rs, the following resources might be interesting:

- [Long term vision document](https://github.com/mxinden/specs/blob/hole-punching/connections/hole-punching.md)
- Status quo hole punching tracking issue across libp2p implementations https://github.com/libp2p/specs/issues/312
- Status quo hole punching tracking issue for rust-libp2p https://github.com/libp2p/rust-libp2p/issues/2052

Let me know in case you need more info. Happy to schedule a call or continue in some async fashion.

elenaf9:
@mxinden in case that you have time (feel free to say no in case that you are busy!), we would really appreciate if we could get some feedback from you on the new interface of our communication library (`communication-refactored/src/interface.rs` / `communication-refactored/src/interface/swarm_task.rs`. 

The general purpose of the interface is to handle the polling of the swarm and provide a convenient way to perform operations on the swarm without blocking.
The interface (`ShCommunication`) is implemented by handling all Swarm interaction in a single task( `SwamTask`). Operations are triggered by sending an according message through a mpsc channel to the swarm-task, and return the outcome via an oneshot channel that is included in the message.  Inbound requests, and general swarm events (/ `NetEvent`) are forwarded through a `mpsc` channel.

It would be great to to know whether you think that our implementation is a valid way to abstract from the underlying rust-libp2p logic, or if you have any concerns with it.

mxinden:
> @mxinden in case that you have time (feel free to say no in case that you are busy!), we would really appreciate if we could get some feedback from you on the new interface of our communication library (communication-refactored/src/interface.rs / communication-refactored/src/interface/swarm_task.rs.

Of course. I will give this pull request a review later this week.

communication-refactored/src/actor.rs
```diff
@@ -0,0 +1,166 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![cfg(feature = "actor")]
+
+pub mod messages;
```
mxinden:
```suggestion
#![cfg(feature = "actor")]
pub mod messages;

```

The `cfg` only applies to the single line, not the whole module, right?

communication-refactored/src/behaviour.rs
```diff
@@ -0,0 +1,1024 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
mxinden: :pray:

communication-refactored/src/behaviour.rs
```diff
@@ -0,0 +1,1024 @@
...
+                    self.firewall.set_rule(peer, inbound, RuleDirection::Inbound)
+                }
+                if outbound.is_some() {
+                    self.firewall.set_rule(peer, outbound, RuleDirection::Outbound)
+                }
+            }
+            self.handle_updated_peer_rule(peer, RuleDirection::Both);
+        }
+
+        // Handle individual approvals fro requests that were returned after a [`FirewallRequest::RequestApproval`]
```
mxinden:
```suggestion
        // Handle individual approvals for requests that were returned after a [`FirewallRequest::RequestApproval`]
```

communication-refactored/src/interface.rs
```diff
@@ -0,0 +1,559 @@
...
+        let _ = self.command_tx.start_send(command);
+    }
+}
+
+/// Use existing keypair for authentication on the transport layer.
+///
+/// The local [`PeerId`] is derived from public key of the IdKeys.
+/// If this is not the case, remote Peers will reject the communication.
+pub enum InitKeypair {
+    /// Identity Keys that are be used to derive the noise keypair and peer id.
```
mxinden:
```suggestion
    /// Identity Keys that are used to derive the noise keypair and peer id.
```

communication-refactored/src/interface.rs
```diff
@@ -0,0 +1,559 @@
...
+            relay_behaviour,
+            self.firewall_channel,
+        );
+        let mut swarm_builder = SwarmBuilder::new(transport, behaviour, peer_id).executor(Box::new(|fut| {
+            tokio::spawn(fut);
+        }));
+        if let Some(limit) = self.connections_limit {
+            swarm_builder = swarm_builder.connection_limits(limit);
+        }
+        let swarm = swarm_builder.build();
```
mxinden:
```suggestion
        let swarm = swarm_builder.substream_upgrade_protocol_override(Version::V1Lazy).build();
```

Consider using `V1Lazy` instead of `V1` to save on a whole network roundtrip during stream negotiation, e.g. when doing a request.

See https://docs.rs/multistream-select/0.10.2/multistream_select/enum.Version.html#variant.V1Lazy for details.

communication-refactored/src/interface.rs
```diff
@@ -0,0 +1,559 @@
...
+{
+    // Id of the local peer.
+    local_peer_id: PeerId,
+    // Channel for sending [`SwarmOperation`] to the [`SwarmTask`] .
+    // The [`SwarmOperation`]s trigger according operations on the Swarm.
+    // The result of an operation is received via the oneshot Receiver that is included in each type.
+    command_tx: mpsc::Sender<SwarmOperation<Rq, Rs, TRq>>,
+}
+
+impl<Rq, Rs, TRq> ShCommunication<Rq, Rs, TRq>
```
mxinden: This looks like a great interface for others to integrate with.

communication-refactored/src/interface/swarm_task.rs
```diff
@@ -0,0 +1,571 @@
...
+                request,
+                response_tx,
+            }) => {
+                let received_rq = ReceiveRequest {
+                    request_id,
+                    peer,
+                    request,
+                    response_tx,
+                };
+                let _ = self.request_tx.try_send(received_rq);
```
mxinden: Is sending these to the _inbound request handler_ on a best-effort basis good enough? Why not properly send them via `SinkExt::send`?

mxinden (commented):
I haven't found the time yet to give this a full review. Submitting an intermediate one for now. I will take another look in the upcoming days.

I like the _Interface_ and _Swarm Task_ abstraction.

communication-refactored/src/actor.rs
elenaf9: It was meant for the whole module. 
But I can actually remove that completely, since I already have in my lib.rs:
```
#[cfg(feature = "actor")]
pub mod actor;
```

communication-refactored/src/interface/swarm_task.rs
elenaf9: But wouldn't `SinktExt::send` require to `await` the returned future, hence potentially block everything? 
And I am not sure how this would behave in case that the `Receiver` is not dropped, but also never read. 

I have considered adding a RingBuffer structure to cache requests in case that `try_send` returns `SendErrorKind::Full`.  In that case the loop in `SwarmTask::run` would also always do something like
```
futures::select! {
      ...
      _ = poll_fn(|cx| request_tx.poll_ready(cx)).fuse() => {
          // send oldest element from ring buffer
      }
  }
```
The benefit from using a ring buffer would be that only the most recent requests would be cached.
Another option would be to use an unbounded channel for inbound requests, but this could potentially use a lot of memory and even be a security risk if a remote is spam-sending requests.

I so far kept it rather simple because usually I'd expect the `Receiver` to be in some sort of loop that continuously reads from it (and if the `Receiver` is dropped there nothing we can do anyway). But I agree that this is probably not good enough.
What do you think about the ring-buffer idea @mxinden?

communication-refactored/src/interface.rs
elenaf9: Thanks for the advice. I do have to admit though that the docs on `V1Lazy` are a bit difficult to understand:
> This strategy is only applicable [...] if the dialer supports just a single application protocol

What does `a single application protocol` mean? That only one version of a protocol is supported?

> Nevertheless, there is a pitfall that is rarely encountered: When nesting multiple protocol negotiations, the listener should either be known to support all of the dialer’s optimistically chosen protocols or there is must be no intermediate protocol without a payload and none of the protocol payloads must have the potential for being mistaken for a multistream-select protocol message

In what cases are protocol negotiations nested? And are there any protocols in rust-libp2p without a payload?

elenaf9:
Thanks @mxinden!

communication-refactored/src/interface.rs
mxinden: > > This strategy is only applicable [...] if the dialer supports just a single application protocol
> 
> What does `a single application protocol` mean? That only one version of a protocol is supported?

Protocols on streams are negotiated using [Multistream Select](https://github.com/multiformats/multistream-select). The process is quite simple. The dialer proposes a protocol, the listener either accepts the protocol by echoing it, or rejects it by sending `na` (not available).

Say the dialer opens a new stream and only has a single protocol that it is willing to speak on that stream. E.g. not `/my-protocol/1` and `/my-protocol/2`, but only a single one, e.g. `/my-protocol/2`. What the dialer can do is send the `/my-protocol/2` and instead of waiting for the echo, or the `na`, directly send application data, not having to wait for 1 round trip.

This only works, when one is willing to speak a single protocol on that stream only. When willing to speak either one of many protocols, `V1Lazy` defaults to the `V1` behavior.


> 
> > Nevertheless, there is a pitfall that is rarely encountered: When nesting multiple protocol negotiations, the listener should either be known to support all of the dialer’s optimistically chosen protocols or there is must be no intermediate protocol without a payload and none of the protocol payloads must have the potential for being mistaken for a multistream-select protocol message
> 
> In what cases are protocol negotiations nested? And are there any protocols in rust-libp2p without a payload?

Negotiations are nested e.g. when first negotiating a security protocol, then, nested, a multiplexer and then, nested a stream protocol. `substream_upgrade_protocol_override(Version::V1Lazy)` only sets the `V1Lazy` option for the last one, not followed by any other nesting, thus safe.

I am not aware of any **intermediate** protocols without a payload.

---

This is just a suggestion. After all, playing it safe by using `V1` is just fine. You can still later on enable `V1Lazy` without having to coordinate a roll out.

communication-refactored/src/interface/swarm_task.rs
mxinden: > But wouldn't SinktExt::send require to await the returned future, hence potentially block everything?

That is correct. You would depend on the other side to continuously read from the channel. Is there reason to worry that the other side will not continuously read from that channel?

> Another option would be to use an unbounded channel for inbound requests, but this could potentially use a lot of memory and even be a security risk if a remote is spam-sending requests.

I would strongly advice against unbounded channels. See for example https://github.com/paritytech/substrate/search?q=unbounded&type=issues for the many issues we had related to unbounded channels.

>  What do you think about the ring-buffer idea @mxinden?

While I think it is a healthy attitude to drop old messages in favor of new ones, I don't quite see the need for the ring buffer in the first place, circling back to my question above, is there reason to worry that the other side will not continuously read from that channel?

communication-refactored/src/interface.rs
```diff
@@ -0,0 +1,559 @@
...
+            request,
+            tx_yield,
+        };
+        self.send_command(command).await;
+        rx_yield.await.unwrap()
+    }
+
+    /// Start listening on the network.
+    /// If no address is given, the listening address will be OS-assigned.
+    pub async fn start_listening(&mut self, address: Option<Multiaddr>) -> Result<Multiaddr, ListenErr> {
```
mxinden: When e.g. listening on `/ip4/0.0.0.0/tcp/0` multiple listen addresses might be returned, no?

communication-refactored/src/interface.rs
elenaf9: Thank you for the detailed explanation, that makes sense. Using `V1Lazy` sounds reasonable.

communication-refactored/src/interface/swarm_task.rs
elenaf9: > Is there reason to worry that the other side will not continuously read from that channel?

In general, I would expect the receiver to continuously read from the channel, yes. 
But I could also imagine use-cases in which a request may require e.g. some asynchronous operations or user interaction, and where requests are handled sequentially. In this case, a next request might only be read from the channel after the preceding one was resolved, which could then potentially take some time. In the meantime, no swarm interaction could take place once the capacity of the channel was reached. I think this behaviour may be desired in some cases, but maybe not in every one.

And if the receiver continuously polls the request channel, `SinktExt::send` would not block, but in this case the current best-effort basis via `Sender::try_send` would also be sufficient, wouldn't it? 

--- 
I guess my general motive is that the `SwarmTask` should not have to rely on any behaviour outside of it's own loop. Because it is running in its own decoupled task, I'd like to guarantee that it will not block, which is why I try to not use any asynchronous operations (apart from the central `run` method).

communication-refactored/src/interface.rs
elenaf9: Yes that's true. The problem is that I don't see a way of determining when the last "new-listen-address" is reported / when I have "collected" all addresses associated with a new listener. I am currently only returning the address from the first reported `SwarmEvent::NewListenAddr` of the associated Listener.
Do you see a way of solving this?

But either way, I will correct my docs, since it currently implies that there is only one listening address, which is not correct in cases like the one you mentioned.

communication-refactored/src/interface.rs
elenaf9: For some reason, `test test_firewall::firewall_ask` does not terminate when using `V1Lazy`, and if disable the feature `tcp-transport` (which results in the use of a plain `TokioTcpConfig` as transport), `test_firewall::firewall_permissions` does not terminate as well. All other tests (dialing, sending requests) pass.

Since I don't really know what the reason for this behaviour could be, I'll use `V1` for now. 
I might investigate this further in the future, but right now it is not a priority.

communication-refactored/src/interface.rs
mxinden: Given that a listener might emit a `NewAddress` at any point in time, the only way I see to properly report this is a channel instead of a oneshot, that lives as long as the listener, being used for each `NewListenAddr`.

That said, I am not sure it is worth the complexity in your use-case.

communication-refactored/src/interface/swarm_task.rs
mxinden: _Take the below with a grain of salt. After all I am not very familiar with your whole stack._

> I think this behaviour may be desired in some cases, 

It would be desired to enforce back-pressure throughout the system. E.g. when the machine is already at its limit, the network should pause for a bit and not produce even more events.

> but maybe not in every one.

I would expect a large enough channel size to cater for those cases.

> And if the receiver continuously polls the request channel, SinktExt::send would not block, but in this case the current best-effort basis via Sender::try_send would also be sufficient, wouldn't it?

It would. Though (a) you loose the first mentioned back-pressure (whether that is useful in your case or not I can't tell) and (b) silently dropped events are hard to debug.

communication-refactored/src/interface.rs
mxinden: >  I might investigate this further in the future, but right now it is not a priority.

In case you get around to it, I would be curious what the cause is.

mxinden:
One note on the usage of Cargo features to enable/disable mdns and relay: Features are additive, thus in case this crate is imported twice with different feature sets, the union of the two will be used.

communication-refactored/src/interface/swarm_task.rs
elenaf9: > It would be desired to enforce back-pressure throughout the system. E.g. when the machine is already at its limit, the network should pause for a bit and not produce even more events.

That's a good point, which I did not consider before.

---
After thinking about this more, I decided to go the extra mile and support all 3 discussed configuration (block/drop/use ring-buffer). I think this behaviour should be decided by the user, especially since Stronghold is only a library and I do not know all use cases.
I have added a new commit https://github.com/iotaledger/stronghold.rs/pull/210/commits/d93c9f632bbcba336ae70e2b18b86e0e286b5111 that implements a wrapper for the `mpsc::Sender` and implements `Sink` for it according to the set config.

p2p/src/interface/msg_channel.rs
```diff
@@ -0,0 +1,180 @@
...
+        match self.inner.try_send(item) {
+            Ok(()) => Ok(()),
+            Err(e) if e.is_full() => match self.buffer.as_mut() {
+                Some(buffer) => {
+                    if buffer.len() >= buffer.capacity() {
+                        // Drop older events in favor of new ones.
+                        buffer.pop_front();
+                    }
+                    buffer.push_back(e.into_inner());
+                    if let Some(waker) = self.waker.as_ref() {
```
mxinden:
```suggestion
                    if let Some(waker) = self.waker.take() {
```

I am not 100% sure, though I would expect to be woken up at most once when waiting for a new item to be added to the `buffer`, no?

elenaf9: I actually had it like this at first, but because I was also not 100% sure I changed it to not consume the waker. 

But you are right. If a new item was added to the buffer and the task was waken again, it will attempt to send the request. Thus wakening the task is then related to the channel being ready, and not to new items in the buffer.

client/Cargo.toml
```diff
@@ -21,8 +21,11 @@ zeroize = "1.1"
 zeroize_derive = "1.0"
 anyhow = "1.0"
 thiserror = "1.0"
-futures = "0.3"
-riker = "0.4"
+futures = { version = "0.3", optional = true }
+actix = "0.12"
+
+# remove dependency
```
felsweg-iota: let's better remove this commented out dependency to riker.

client/src/interface.rs
```diff
@@ -486,93 +577,87 @@ impl Stronghold {
...
 impl Stronghold {
-    /// Spawn the communication actor and swarm with a pre-existing keypair
-    /// Per default, the firewall allows all outgoing, and reject all incoming requests.
-    pub fn spawn_communication_with_keypair(&mut self, keypair: Keypair) -> StatusMessage {
-        if self.communication_actor.is_some() {
-            return StatusMessage::Error(String::from("Communication was already spawned"));
+    /// Spawn the p2p-network actor and swarm.
+    pub async fn spawn_p2p(&mut self, firewall_rule: Rule<ShRequest>, network_config: NetworkConfig) -> StatusMessage {
+        if self.network_actor.is_some() {
+            return StatusMessage::Error(String::from("Network actor was already spawned"));
```
felsweg-iota, line 568: We can think of further extending the error types, or even having a `P2PError`. Would be good to handle certain errors explicitly.

client/src/interface.rs
```diff
@@ -486,93 +577,87 @@ impl Stronghold {
...
-
-        let communication_actor = self
-            .system
-            .actor_of_args::<CommunicationActor<_, SHResults, _, _>, _>(
-                "communication",
-                (keypair, actor_config, behaviour_config),
-            )
-            .expect(line_error!());
-        self.communication_actor = Some(communication_actor);
+        let network_actor = unwrap_or_err!(NetworkActor::new(self.target.clone(), firewall_rule, network_config).await);
```
felsweg-iota, line 570: Let's think on moving the convenience macros to crate level.

client/src/interface.rs
```diff
@@ -486,93 +577,87 @@ impl Stronghold {
...
+                Err(_e) => (empty_response, StatusMessage::Error("No payload present".into())),
+            },
+            Err(_e) => (empty_response, StatusMessage::Error("No secret present".into())),
         }
     }
 }
 
-#[cfg(feature = "communication")]
+#[macro_export]
+macro_rules! unwrap_or_err (
```
felsweg-iota: Let's think of using error types rather than strings.

client/src/interface.rs
```diff
@@ -585,324 +670,227 @@ impl Stronghold {
...
+        let actor = match self.network_actor.as_ref() {
+            Some(a) => a,
+            None => return ProcResult::Error("No network actor spawned.".to_string()),
+        };
+        let send_request = network_msg::SendRequest {
+            peer,
+            request: CallProcedure { proc: control_request },
+        };
+        match actor.send(send_request).await {
+            Ok(Ok(Ok(res))) => res,
```
felsweg-iota: hmm is there a way to unwrap it in a more clear way? This nesting of results seem redundant, maybe we can work around that?

p2p/src/behaviour.rs
```diff
@@ -0,0 +1,1031 @@
...
+        self.addresses.add_relay(peer, address)
+    }
+
+    #[cfg(feature = "relay")]
+    /// Remove a relay from the list of dialing relays.
+    pub fn remove_dialing_relay(&mut self, peer: &PeerId) {
+        self.addresses.remove_relay(peer);
+    }
+
+    #[cfg(feature = "relay")]
```
felsweg-iota: we can think of moving all feature gates parts of the code into an own module.

p2p/src/behaviour.rs
```diff
@@ -0,0 +1,1031 @@
...
+            .and_then(move |()| rule_rx.map_err(|_| ()))
+            .map_ok_or_else(move |()| (peer, None), move |rules| (peer, Some(rules)))
+            .boxed();
+        self.pending_rule_rqs.push(future);
+        self.request_manager.add_pending_rule_request(peer);
+    }
+
+    // Query for individual approval of a requests.
+    // This is necessary if the firewall is configured with [`Rule::Ask`].
+    fn query_rq_approval(&mut self, peer: PeerId, request_id: RequestId, rq: TRq, direction: RequestDirection) {
```
felsweg-iota: Better use words eg. `query_request_approval`

p2p/src/behaviour.rs
```diff
@@ -0,0 +1,1031 @@
...
+                }
+            };
+            return Poll::Ready(action);
+        }
+        Poll::Pending
+    }
+}
+
+#[cfg(test)]
+mod test {
```
felsweg-iota, line 799: Better move tests outside the module.

p2p/src/behaviour/firewall/permissions.rs
```diff
@@ -0,0 +1,94 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+pub use stronghold_derive::RequestPermissions;
+
+/// The permission value for request variants.
+/// This is realized as a bit set at a certain index, hence the value is always a power of 2.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct PermissionValue(u32);
+
+impl PermissionValue {
```
felsweg-iota, line 11: This could be generic, that `PermissionValue` would have generic Type `T`, that is being comparable and hashable. This way we can keep the number of permissions dynamic. I can't see a reason why the permissions have a purely numeric value that is being stored inside an unsigned int. Does the inherent ordering have a semantic value inside the firewall rules?

p2p/src/behaviour/request_manager.rs
```diff
@@ -0,0 +1,573 @@
...
+    PeerId,
+};
+use smallvec::SmallVec;
+use std::{
+    borrow::Borrow,
+    collections::{HashMap, VecDeque},
+    marker::PhantomData,
+};
+
+// Actions for the behaviour to handle i.g. the behaviour emits the appropriate `NetworkBehaviourAction`.
```
felsweg-iota, line 24:
```suggestion
// Actions for the behaviour to handle eg. the behaviour emits the appropriate `NetworkBehaviourAction`.
```

felsweg-iota (commented):
That was a big PR and a good one. Good work! From my side the code is good so far.

client/src/interface.rs
elenaf9: I agree, but at the moment it is returning a String for the Error in order to be consistent with the return values of the other interface methods. 
I think we should do a separate PR, that changes the general interface to return `anyhow::Error`s in the `ResulMessage::Error`, and add a P2PError-type in that PR. 
Thoughts @felsweg-iota ?

client/src/interface.rs
elenaf9: See comment above.

p2p/src/behaviour.rs
elenaf9: This is not possible here, as it is testing the `NetBehaviour`, which is not exposed to the outside of the crate.

p2p/src/behaviour/firewall/permissions.rs
elenaf9: In general, the logic inside of this `permissions.rs` module is not used anymore in the stronghold-p2p library itself. Instead it provides some logic for convenience, that may be used by the user when writing a `Rule::Restricted. An example for this implemented in the test `test_firewall::firewall_permissions`.

But it being a numeric value is on purpose yes. The idea is (similar to how the `rwx` permissions in unix systems implement it) that each variant has a numeric value to the power of 2, e.g. ReadStore: 1, WriteStore: 2, WriteVault: 4, ExecuteProcedure: 8, ..., so that the a `FirewallPermission` can be created from the sum of allowed `PermissionValue`s.
It could be implemented that the type of integer is generic, but I decided against it because 1. It's unlikely that there are request messages enum with more than 32 variants , and 2. the user can always do their own implementation if this doesn't fit their use-case.

client/src/interface.rs
elenaf9: Ah I just noticed that with the recent changes in the interface, it already returns an anyhow Error.
I will add the P2PError then and bubble up the actual error types of the p2p library.

lucas-tortora (approved):
LGTM!

client/src/interface.rs
```diff
@@ -843,20 +820,17 @@ impl Stronghold {
...
         let send_request = network_msg::SendRequest {
             peer,
             request: CallProcedure { proc: control_request },
         };
-        match actor.send(send_request).await {
-            Ok(Ok(Ok(res))) => res,
-            Ok(Ok(Err(e))) => ProcResult::Error(e.to_string()),
-            Ok(Err(e)) => ProcResult::Error(e.to_string()),
-            Err(e) => ProcResult::Error(e.to_string()),
+        let receive_response = unwrap_or_err!(actor.send(send_request).await);
```
felsweg-iota, line 829: nice!

felsweg-iota (approved):
good one.

client/src/interface.rs
felsweg-iota: I agree on specific error types, where they make sense to provide a more consistent way of handling them.

p2p/src/behaviour/firewall/permissions.rs
felsweg-iota: I do see the point, and would leave it. In general I would advise to always be open towards more options and flexibility.

PR #222, opened 2021-06-29
fix(comms): add / fix stop listening

elenaf9:
# Description of change

Add method `stop_listening` to `Stronghold` interface.
Fix stop-listening in `SwarmTask` in communication.


felsweg-iota (approved):
looks good.

PR #243, opened 2021-08-10
Feat/actix migration

felsweg-iota:
# Description of change

Migrating actor system from [riker](https://riker.rs) to [actix](https://actix.rs). Certain features from previous implementation have been merged. The former client and internal are now merged into one. Message types for actors are now separated as structs. The interface to stronghold has not changed, but the initialization of stronghold is now asynchronous. The current migration does not pull in the communications features for stronghold. 

none.

## Type of change

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist


client/Cargo.toml
```diff
@@ -21,8 +21,11 @@ zeroize = "1.1"
 zeroize_derive = "1.0"
 anyhow = "1.0"
 thiserror = "1.0"
-futures = "0.3"
-riker = "0.4"
+# futures = "0.3"
+actix = "0.12"
+
+# remove dependency
+# riker = "0.4"
```
elenaf9:
```suggestion
```

client/README.md
```diff
@@ -1,6 +1,6 @@
 ## Stronghold Client
 
-This is the official client layer of Stronghold. It uses the Riker actor model and provides an easy interface which acts as a functional passthrough to Stronghold's internal actor system. 
+This is the official client layer of Stronghold. It provides an [Actix](https://actix.rs) actor model system for easy Interface as well as functional passthrough to Stronghold's internal actor system for integrators not using Riker.
```
elenaf9:
```suggestion
This is the official client layer of Stronghold. It provides an [Actix](https://actix.rs) actor model system for easy Interface as well as functional passthrough to Stronghold's internal actor system.
```
Riker is not relevant for Stronghold anymore, hence there is no need to mention it. Apart from that, it may confuse newer users that don't know about the past usage of Riker.

client/examples/cli/main.rs
```diff
@@ -5,10 +5,10 @@ mod arguments;
 
 use arguments::*;
 use clap::Clap;
-use futures::executor::block_on;
+// use futures::executor::block_on;
 use iota_stronghold::{home_dir, naive_kdf, Location, RecordHint, StatusMessage, Stronghold};
 
-use riker::actors::*;
+// use riker::actors::*;
```
elenaf9: remove dead code

client/src/actors.rs
```diff
@@ -1,15 +1,24 @@
 // Copyright 2020-2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-mod client;
-mod internal;
+mod registry;
+mod secure;
```
elenaf9: I think calling this module `secure` is misleading, as the content of this module does not really introduce any security functionality. Consider renaming it back to `client`.

client/src/actors/secure.rs
```diff
@@ -0,0 +1,1048 @@
...
+    let vids = keystore.keys().copied().collect::<HashSet<VaultId>>();
+    self.keystore.rebuild_keystore(keystore);
+    self.db = state;
+    self.rebuild_cache(self.client_id, vids, store);
+});
+
+impl_handler!(messages::CheckVault, Result<(), anyhow::Error>, (self, msg, _ctx), {
+    let vid = self.derive_vault_id(msg.vault_path);
+    self.vault_exist(vid).ok_or(anyhow::anyhow!(VaultError::NotExisting)).map(|_|())
+});
```
elenaf9: I would expect `CheckVault`, analogous to `CheckRecord`, to return a boolean. The status that the Vault does not exist, is a completely acceptable / "ok" state. 
In my opinion, Errors should only be returned on actual error-nous cases, e.g. if a prerequisite for this operation was not fulfilled, if the user attempts to do something forbidden, if there was an unexepected failure, etc.

client/src/actors/snapshot.rs
```diff
@@ -3,149 +3,156 @@
...
+/// re-export local modules
+pub use messages::*;
+pub use returntypes::*;
+
+pub mod returntypes {
+
+    use super::*;
+
+    /// Return type for loaded snapshot file
+    pub struct ReturnReadSnapshot<T: BoxProvider + Send + Sync + Clone + 'static + Unpin> {
```
elenaf9: Using the Generic `T` here is redundant and doesn't bring any benefit, since line 110 is restricting this Generic to `internals::Provider`. There is no way that a different provider can be used. I would suggest to remove the Generic and use directly `internals::Provider` here.

client/src/interface.rs
```diff
@@ -1,90 +1,94 @@
...
-
-        clients.insert(client_id, client.clone());
-
-        Self {
-            system,
-            clients,
+    /// Initializes a new instance of the system.  Sets up the first client actor. Accepts an optional [`SystemRunner`],
+    /// the first client_path: `Vec<u8>` and any `StrongholdFlags` which pertain to the first actor.
+    /// - The [`SystemRunner`] is not being used directly by stronghold, but is being initialized on the first run.
+    /// - The initialization function can be made asynchronous as well, getting rid of internal explicit blocking
```
elenaf9:
```suggestion
    /// Initializes a new instance of the system.  Sets up the first client actor. Accepts the first client_path:
    /// `Vec<u8>` and any `StrongholdFlags` which pertain to the first actor.
```
Fix (outdated ?) doc, since there is no `SystemRunner` parameter here.

client/src/interface.rs
```diff
@@ -219,21 +209,20 @@ impl Stronghold {
         payload: Vec<u8>,
         lifetime: Option<Duration>,
     ) -> StatusMessage {
-        let res: SHResults = ask(
-            &self.system,
-            &self.target,
-            SHRequest::WriteToStore {
+        // TODO move to top
```
elenaf9: Fix TODOs here and in other methods: move the imports to the top. 
Especially since those message will be used in the remote-execution mirror of theses methods (e.g. `write_remote_store`) as well, it would make the code much cleaner.

client/src/interface.rs
```diff
@@ -242,76 +231,82 @@ impl Stronghold {
...
-        if let SHResults::ReturnGarbage(status) =
-            ask(&self.system, &self.target, SHRequest::GarbageCollect(vault_path)).await
+        use crate::actors::secure_messages::GarbageCollect;
+
+        match self
+            .target
+            .send(GarbageCollect {
+                location: Location::Generic {
+                    vault_path,
+                    record_path: Vec::new(), // this will be dropped.
```
elenaf9: Why use Location instead of directly VaultId? This is redundant, as it creates a location and then in the SecureClient parses it back to the vault_id.

client/src/lib.rs
```diff
@@ -5,14 +5,15 @@
...
-/// An interface for implementing the stronghold engine. Using the Riker Actor model, this library provides a
+/// An interface for implementing the stronghold engine. Using the Actix Actor model, this library provides a
 /// mechanism to manage secret data between multiple users. Stronghold may be accessed via the `Stronghold`
 /// object. The interface contains methods to access the secure runtime environment and methods to write to the
 /// Stronghold. Each Stronghold contains a collection of versioned records, identified as Vaults. Each Vault
 /// contains a set of versioned records of like data. Multiple clients can be spawned with Stronghold, each of
 /// which can hold multiple vaults (See the `Location` API for more details). The Stronghold interface also
 /// contains a generic insecure key/value store which can be accessed as a `Store`. Each client contains a single
 /// store and the same location may be used across multiple clients.
+// TODO: Adapt Documentation
```
elenaf9: Fix Todo.

client/src/state/secure.rs
```diff
@@ -1,39 +1,42 @@
...
+/// Cache type definition
 pub type Store = Cache<Vec<u8>, Vec<u8>>;
 
-/// A [`Client`] Cache Actor which routes external messages to the rest of the Stronghold system.
-#[actor(SHResults, SHRequest, InternalResults)]
-#[derive(Debug, Serialize, Deserialize, Clone)]
-pub struct Client {
+pub struct SecureClient<Pr>
+where
+    Pr: BoxProvider + Send + Sync + Clone + 'static + Unpin,
```
elenaf9: Remove the Generic, instead directly use `internals::Provider`, since this is redundant. The interface, the Registry and the snapshot actor all replace this generic with  `internals::Provider`, thus there is no way to use the `SecureClient` with any other Provider.

client/src/state/snapshot.rs
```diff
@@ -10,13 +10,13 @@ use engine::{
...
 
-use crate::{line_error, state::client::Store, Provider};
+use crate::{line_error, state::secure::Store, Provider};
 
 use std::path::Path;
 
 use std::collections::HashMap;
 
-/// Wrapper for the [`SnapshotState`] data structure.
+#[derive(Default)]
```
elenaf9:
```suggestion
/// Wrapper for the [`SnapshotState`] data structure.
#[derive(Default)]
```

client/src/actors/registry.rs
```diff
@@ -0,0 +1,229 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Registry Actor
+//!
+//! The registry actor keeps record of all client actors
+//! bound to a unique `client_id`. [`Client`] actors can
```
elenaf9:
```suggestion
//! bound to a unique `client_id`. [`SecureClient`] actors can
```
Check & fix intra doc links:
```sh 
$ RUSTDOCFLAGS="--deny broken_intra_doc_links" cargo doc --verbose --workspace --no-deps --document-private-items
```

client/src/state/secure.rs
felsweg-iota: See similar comment above.

client/src/actors.rs
felsweg-iota: The secure module contains the actor that is responsible for accessing the secure features inside stronghold, so I would argue the name make sense. The `SecureActor` renders the `ClientActor` and `InternalActor` obsolete, while providing the same level of isolation.

client/src/actors/snapshot.rs
felsweg-iota: Good point on code consistency. The `engine::vault::crypto_box::BoxProvider` trait is open, so we can assume alternative implementation. But for now we can stick to the current one.

client/src/actors/secure.rs
felsweg-iota: Depends on how you interpret the error. The explicit `VaultError::NotExisting` can be handled within a match, while a bool would semantically be different. A message type called `IsVaultExisting` would be boolean typed. But we can argue, that the current implementation only allows the return of an error indicating the absence of a vault. We would have two options here: 
- introduce a more explicit message type called `isVaultExisting`; return a boolean type for this message handler as return type
- extend `CheckVault`, that other return values are possible aside from checking the existence of a vault.

client/src/interface.rs
felsweg-iota: no entirely correct. `GarbageCollect` receives a `location`, that will be resolved to a `vault_id`. The `Location` type is being used to use the existing infrastructure code. 

in [secure.rs](https://github.com/iotaledger/stronghold.rs/blob/44b06196a522b458db52f3af3210398c0c3e13fd/client/src/actors/secure.rs#L523) we have
```
let (vault_id, _) = self.resolve_location(msg.location);
```
that is different, then just packing `vault_path` into a `Location` and the simply unpacking it again.

client/src/actors.rs
tensor-programming: Hmm, honestly it could go either way.  Since none of this is being exposed through the API it doesn't really matter except to people who are working on stronghold internally.

client/src/actors/registry.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn handle(&mut self, _: messages::GetAllClients, _: &mut Self::Context) -> Self::Result {
+        let mut result = Vec::new();
+
+        for (id, addr) in &self.clients {
+            result.push((*id, addr.clone()));
+        }
+        result
+    }
+}
+
```
tensor-programming: maybe move these tests out of here.

client/src/state/secure.rs
```diff
@@ -212,4 +208,26 @@ mod test {
         assert_eq!(test_rid, rid);
         assert_eq!(2, ctr);
     }
+
+    #[test]
```
tensor-programming: again probably worth moving the tests out.

client/src/tests/procedures_tests.rs
```diff
@@ -3,24 +3,25 @@
 
 #![allow(non_snake_case)]
 
-use riker::actors::ActorSystem;
+// use actix::Actor;
+// use riker::actors::ActorSystem;
```
tensor-programming: Cleanup the comments from the old imports.  No need to keep the riker stuff since its gone.

tensor-programming (approved):
Overall, looks like a good refactor.  I am still slightly skeptical on having everything in a single actor, perhaps in the future consider looking into some kind of dynamic "runtime" actor for running procs at least.    Would be curious to see how an audit would go over with this current implementation (not that I think there would be any security problems here).

nothingismagick (approved):
I think that this PR is fantastic. Great job @felsweg-iota

felsweg-iota:
> Overall, looks like a good refactor. I am still slightly skeptical on having everything in a single actor, perhaps in the future consider looking into some kind of dynamic "runtime" actor for running procs at least. Would be curious to see how an audit would go over with this current implementation (not that I think there would be any security problems here).

I agree on future refactoring of the current secure actor, and split it up alongside the upcoming procedures api.

PR #251, opened 2021-08-26
fix: inconsistent error message.

felsweg-iota:
# Description of change

This fixes an inconsistent error message when trying to access a store location.

none. 

## Type of change

unit and integration tests

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist

### DIFF
--- a/client/src/actors/secure.rs
+++ b/client/src/actors/secure.rs
@@ -66,6 +66,12 @@ pub enum VaultError {
 }
 
 #[derive(DeriveError, Debug)]
+pub enum StoreError {
+    #[error("Unable to read from store")]
+    NotExisting,
+}
+
+#[derive(DeriveError, Debug)]
 pub enum SnapshotError {
     #[error("No snapshot present for client id ({0})")]
     NoSnapshotPresent(String),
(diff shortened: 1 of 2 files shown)


nothingismagick (changes requested):
This makes good sense. Please add a .changes file for this :)

nothingismagick (approved):

client/src/actors/secure.rs
```diff
@@ -65,6 +65,12 @@ pub enum VaultError {
     AccessError,
 }
 
+#[derive(DeriveError, Debug)]
+pub enum StoreError {
+    #[error("Unable to read from store")]
```
elenaf9, line 70:
```suggestion
    #[error("The store does not exist")]
```
Wouldn't we want to provide the info **why** it is _"Unable to read from store"_, ig. that is does not exist?

felsweg-iota: This fix re-integrates the error message from before, and restores some "expected behavior"*. I would suggest to work on more proper error types. Eventually increasing the cognitive value of this one. So, I do agree. 

* on a functional level the behavior itself has not changed, but the former error message created some confusion, hence this quick fix.

elenaf9 (approved):
LGTM

PR #254, opened 2021-09-03
Fix(client): Avoid Key cloning, remove redundant code

elenaf9:
# Description of change

Adjust the handling of `Key`s in the `SecureClient` and remove obsolete code.

https://github.com/iotaledger/stronghold.rs/commit/da621b74944aedfa652194d4b51a8ea757a68d98:
- Avoid cloning of `Key`s by changing the key-handling in the `SecureClient` to follow the flow:
  _Take key from keystore -> Use key -> Insert key back into keystore_
- Rename `get_key` to `take_key`
- directly use concrete type `internals::Provider` in the `KeyStore` 

https://github.com/iotaledger/stronghold.rs/commit/e2e2fb93f126096ab2b38f7ce8176a64a87e6744:

- Remove `vaults: HashSet<VaultId>` property from the `SecureClient`. 
Before #248, this property was necessary in the `Client` because it did no have access to the `KeyStore`. Since the Keystore is now part of the `SecureClient`, the info about existing vaults can be gathered from the keystore, which makes the VaultId-Hashset obsolete.
- `CheckVault` is now checking `Keystore::vault_exists` and returns a `bool` instead of a `Result`.

## Type of change

Tested via existing tests.

## Change checklist

### DIFF
--- a/.changes/key-cloning.md
+++ b/.changes/key-cloning.md
@@ -3,5 +3,5 @@
 ---
 
 [[PR 254](https://github.com/iotaledger/stronghold.rs/pull/254)]  
-Change key handling in the `SecureClient` to avoid unnecessary cloning of `Key`s.
+Change key handling in the `SecureClient` to avoid unnecessary cloning of keys.
 Remove obsolete VaultId-HashSet from the `SecureClient`.
--- a/.changes/key-cloning.md
+++ b/.changes/key-cloning.md
@@ -1,0 +1,7 @@
+---
+"iota-stronghold": patch
+---
+
+[[PR 254](https://github.com/iotaledger/stronghold.rs/pull/254)]  
+Change key handling in the `SecureClient` to avoid unnecessary cloning of `Key`s.
+Remove obsolete VaultId-HashSet from the `SecureClient`.
--- a/client/src/actors/secure.rs
+++ b/client/src/actors/secure.rs
@@ -835,7 +835,13 @@ impl_handler!(procedures::SLIP10Generate, Result<crate::ProcResult, anyhow::Erro
     let key = self.keystore.take_key(msg.vault_id).unwrap();
 
     let mut seed = vec![0u8; msg.size_bytes];
-    fill(&mut seed).map_err(|e| anyhow::anyhow!(e))?;
+    match fill(&mut seed) {
+        Ok(_) => {},
+        Err(e) => {
+            self.keystore.insert_key(msg.vault_id, key);
+            return Err(anyhow::anyhow!(e))
+        }
+    }
 
     let res = self.db.write(&key, msg.vault_id, msg.record_id, &seed, msg.hint);
 
@@ -855,7 +861,13 @@ impl_handler!(procedures::SLIP10DeriveFromSeed, Result<crate::ProcResult, anyhow
 
     if !self.keystore.vault_exists(msg.key_vault_id) {
         let key = self.keystore.create_key(msg.key_vault_id);
-        self.db.init_vault(key, msg.key_vault_id)?;
+        match self.db.init_vault(key, msg.key_vault_id) {
+            Ok(_) => {},
+            Err(e) => {
+                self.keystore.insert_key(msg.seed_vault_id, seed_key);
+                return Err(anyhow::anyhow!(e))
+            }
+        }
     }
     let dk_key = self.keystore.take_key(msg.key_vault_id).unwrap();
 
@@ -903,7 +915,13 @@ impl_handler!( procedures::SLIP10DeriveFromKey,Result<crate::ProcResult, anyhow:
 
     if !self.keystore.vault_exists(msg.child_vault_id) {
         let key = self.keystore.create_key(msg.child_vault_id);
-        self.db.init_vault(key, msg.child_vault_id)?;
+        match self.db.init_vault(key, msg.child_vault_id) {
+            Ok(_) => {},
+            Err(e) => {
+                self.keystore.insert_key(msg.parent_vault_id, parent_key);
+                return Err(anyhow::anyhow!(e))
+            }
+        }
     }
     let child_key = self.keystore.take_key(msg.child_vault_id).unwrap();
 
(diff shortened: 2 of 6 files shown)


.changes/key-cloning.md
```diff
@@ -0,0 +1,7 @@
+---
+"iota-stronghold": patch
+---
+
+[[PR 254](https://github.com/iotaledger/stronghold.rs/pull/254)]  
+Change key handling in the `SecureClient` to avoid unnecessary cloning of `Key`s.
```
felsweg-iota:
```suggestion
Change key handling in the `SecureClient` to avoid unnecessary cloning of keys.
```

client/src/state/key_store.rs
```diff
@@ -1,24 +1,24 @@
...
+impl KeyStore {
     /// Creates a new [`KeyStore`].
     pub fn new() -> Self {
         Self { store: HashMap::new() }
     }
 
-    /// Gets the key from the [`KeyStore`] and removes it.  Returns an [`Option<Key<P>>`]
-    pub fn get_key(&mut self, id: VaultId) -> Option<Key<P>> {
+    /// Gets the key from the [`KeyStore`] and removes it.  Returns an [`Option<Key<Provider>>`]
+    pub fn take_key(&mut self, id: VaultId) -> Option<Key<Provider>> {
```
felsweg-iota: good change. makes the intent more clear.

client/src/actors/secure.rs
```diff
@@ -842,277 +827,260 @@ impl Handler<CallProcedure> for SecureClient {
...
-    fill(&mut seed).map_err(|e| anyhow::anyhow!(e))?;
+    match fill(&mut seed) {
+        Ok(_) => {},
+        Err(e) => {
+            self.keystore.insert_key(msg.vault_id, key);
+            return Err(anyhow::anyhow!(e))
+        }
+    }
 
-    match self.db.write(&key, msg.vault_id, msg.record_id,&seed, msg.hint).map_err(|e| anyhow::anyhow!(e)) {
```
felsweg-iota, line 861: good idea to get reduce the number of `anyhow` error conversions.

client/src/actors/secure.rs
```diff
@@ -842,277 +827,260 @@ impl Handler<CallProcedure> for SecureClient {
...
-            match self.db
-                .get_guard(&pkey, msg.vault_id, msg.record_id, |data| {
-                    let raw = data.borrow();
-                    let mut raw = (*raw).to_vec();
-
-                    if raw.len() < 32 {
+    let pkey = self
+        .keystore
+        .take_key(msg.vault_id)
+        .ok_or(anyhow::anyhow!(VaultError::NotExisting))?;
```
felsweg-iota: Maybe better directly return the error type.

felsweg-iota (commented):
definitely a good quality of life update. improves the understanding of the code even more. gets rid of redundant code.

client/src/actors/secure.rs
tensor-programming: agreed.

client/src/state/secure.rs
```diff
@@ -11,34 +11,29 @@ use engine::{
...
-use std::{collections::HashSet, time::Duration};
+use std::time::Duration;
 
 /// Cache type definition
 pub type Store = Cache<Vec<u8>, Vec<u8>>;
 
 pub struct SecureClient {
     // A keystore
-    pub(crate) keystore: KeyStore<internals::Provider>,
+    pub(crate) keystore: KeyStore,
```
tensor-programming, line 21: might be a breaking change for the snapshot data.  Would recommend checking to see if the bincode representation of the `KeyStore` is the same as `KeyStore<P>`.

client/src/actors/secure.rs
```diff
@@ -842,277 +827,260 @@ impl Handler<CallProcedure> for SecureClient {
...
-                    let result = result.get();
-                    Ok(ProcResult::SLIP10Derive(ResultMessage::Ok(result)))
-                }
-                Err(e) => {
-                    Err(anyhow::anyhow!(e))
-                }
+    let seed_key = self
+        .keystore
+        .take_key(msg.seed_vault_id)
+        .ok_or(anyhow::anyhow!(VaultError::NotExisting))?;
```
tensor-programming: Seems like a very round-about way to return an error.  Might be better to use `map_err` in the `take_key` function and then just return the error directly that way with 
```
let seed_key = self
        .keystore
        .take_key(msg.seed_vault_id)?
```

Would require changing the `take_key` return to an `Result` type but might be useful given how many times you do this conversion.   See how often its the `Option` type is being used in the context of the rest of the code to weigh whether or not this change is worth it overall.

felsweg-iota (approved):

tensor-programming (changes requested):

tensor-programming (approved):

client/src/state/secure.rs
elenaf9: I don't think it changes anything for the snapshot data. 
The messages that gather the Information from the `SecureClient` (`GetData`) and send it to the `SnapshotActoor` (`FillSnapshot`) do not include the `Keystore`, but instead only the `Key` and the `DbView`:
```
pub struct GetData {}
impl Message for GetData {
    type Result = Result<
        Box<(
            HashMap<VaultId, Key<internals::Provider>>,
            DbView<internals::Provider>,
            Store,
        )>,
        anyhow::Error,
    >;
}

pub struct FillSnapshot {
    pub data: Box<(HashMap<VaultId, Key<Provider>>, DbView<Provider>, Store)>,
    pub id: ClientId,
}
```
Since they already used the concrete `internals::Provider` and were not changed in this PR, I don't think it changed anything for the snapshot.
Please correct me in case that I am missing something. I agree that this is something we should be very careful with.

tensor-programming: Correct.  It should be ok since the state uses the provider directly in there and also doesn't use the keystore.   Could have sworn that we just serialized the keystore as is.  But I guess that was in a much earlier iteration.

PR #258, opened 2021-09-16
Procedures API

elenaf9:
# Description of change

This PR is drafting a new API for Strongholds procedures. The goal is to add more procedures to Stronghold, based on the cryptographic primitives in [crypto.rs](https://github.com/iotaledger/crypto.rs), and to allow the user to create complex procedures that consist of multiple steps. This is especially useful for the execution of procedures on a remote Stronghold, since a chain of multiple procedures like 
```
Slip10Generate -> Slip10Derive -> Ed25519Sign
```
can then be executed through a single request.

###  Concept

_See [this HackMd](https://hackmd.io/-LIn984FQ0CmZroSiBz5mg) for detailed description._

Procedures are classified into 4 different types, depending on whether they use an existing secret and/ or create a new secret that is written into the Vault:
- `GenerateSecret`: no existing secret used, new secret created. e.g. `Slip10Generate`
- `DeriveSecret`: use existing secret, create new secret. e.g. `Slip10Derive`
- `UseSecret`: use existing secret, no new secret created. e.g. `Eded5519Sign`
- `ProcessOutput`: no secret used, no secret created. e.g. `SHA256Digest`

Apart from that, Procedures may have plain input and output, e.g.the input message & signed output in case of `Ed25519Sign`.
The output of each Procedure (if there is an output) is written into a HashMap (=`CollectedOutput`), that is passed to the subsequent procedures and returned to the user in the end. For each Procedure that has plain input, it may be set dynamically whether the procedure should have a pre-set input, or if it gathers its input from the `CollectedOutput` (= uses the output of a preceding procedure). Analogous to this, `DeriveSecret`-/`UseSecret`-Procedures may also use either existing secrets, or secrets that are created in a preceding procedure-step during execution.
Products of a procedure (new secret /  output data) are only temporary if not specified otherwise via `Procedure::write_secret(..)`/ `Procedure::store_output(..)`. Temporary secrets (= records in the vault) are revoked when the procedure returns, and temporary output data is removed from the `CollectedOutput` before returning. 
In the below code example, the seed created in `Slip10Generate` would only be temporary, whereas for ` Slip10Derive ` the new private key would stay in the vault and the output chain code is written into the `CollectedOutput` and returned to the user.
```rust
let sh = Stronghold::init_stronghold_system("my-client".into(), vec![]).await?;

// Create a new seed, that only temporary exists during execution time
let generate = Slip10Generate::default();

let chain = Chain::from_u32(vec![0]);
// Use the newly created seed as input
let derive = Slip10Derive::new_from_seed(generate.target(), chain);

// Identifier / Key for a procedure's output
let k = OutputKey::new("chain-code");

// Target vault + record, in which the derived key will be written
let private_key = Location::generic("my-vault", "child-key");

let hint = RecordHint::new("private-key").unwrap();
let derive = derive
     .store_output(k) // Configure that the output should be returned after execution
     .write_secret(private_key, hint); // Set the record to be permanent in the vault

// Execute the procedures
let output = sh.runtime_exec(generate.then(derive)).await??;

// Get the desired output
let chain_code: ChainCode = output.single_output().unwrap();
```
__Note:__ Procedures are treated as _atomic_: if a single procedure-step in the chain errors, all newly created records are revoked and garbage collected, the collected output is dropped and the error is returned.

**Crypto.rs:**
For the cryptographic primitives some traits/ structs have been added to abstract from the concrete implementations. For now, they are rather basic and only part of the procedures API. However, it can be discussed whether some of these changes should be also added to crypto.rs, to have a unified interface for algorithms of the same type (at the current point in time, crypto.rs only defines traits for the Aead algorithms). 
Some first ideas are written in [this HackMd](https://hackmd.io/X5OCQIGdS9e6Eq9cpCnO1g).

## Type of change

Added new procedures tests.

## Change checklist


client/src/interface.rs
```diff
@@ -822,19 +823,18 @@ impl Stronghold {
...
             peer,
-            request: CallProcedure { proc: control_request },
+            request: control_request.into(),
         };
-        let receive_response = unwrap_or_err!(actor.send(send_request).await);
-        let result = unwrap_or_err!(receive_response);
-        match result {
-            Ok(ok) => ResultMessage::Ok(ok),
-            Err(err) => ResultMessage::Error(err.to_string()),
+        match unwrap_result_msg!(actor.send(send_request).await) {
```
felsweg-iota: Good idea with macro, but what happens when a [MailboxError](https://docs.rs/actix/0.12.0/actix/enum.MailboxError.html) occurs?

client/src/procedures.rs
```diff
@@ -0,0 +1,440 @@
...
+impl State {
+    pub fn insert_data(&mut self, key: OutputKey, value: ProcedureIo, is_temp: bool) {
+        self.aggregated_output.insert(key, (value, is_temp));
+    }
+
+    pub fn get_data(&self, key: &OutputKey) -> Result<&ProcedureIo, anyhow::Error> {
+        self.aggregated_output
+            .get(key)
+            .map(|(data, _)| data)
+            .ok_or_else(|| anyhow::anyhow!("Missing Data"))
```
felsweg-iota: Better use specific error types eg `ProcedureError::MissingData` or similar.

client/src/procedures.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+}
+
+impl From<Vec<u8>> for ProcedureIo {
+    fn from(v: Vec<u8>) -> Self {
+        ProcedureIo(v)
+    }
+}
+
+impl From<ProcedureIo> for Vec<u8> {
```
felsweg-iota: The implementation for `From` for `ProcedureIO` is complemented with the automatic implementation of `Into` for the target type `Vec<u8>`, so no need here to implement it again :)

client/src/procedures.rs
```diff
@@ -0,0 +1,440 @@
...
+// ==========================
+//  Traits for the `Procedure` derive-macro
+// ==========================
+
+pub struct Products<T> {
+    pub secret: Vec<u8>,
+    pub output: T,
+}
+
+trait Parse {
```
felsweg-iota: Better use nouns for traits or types, but that's minimal

client/src/procedures.rs
```diff
@@ -0,0 +1,440 @@
...
+
+impl Target {
+    pub fn random() -> Self {
+        let location = Location::generic(Self::random_path(), Self::random_path());
+        let hint = RecordHint::new("".to_string()).unwrap();
+        Target { location, hint }
+    }
+
+    fn random_path() -> impl Into<Vec<u8>> {
+        let l = (rand::random::<u8>() % (u8::MAX - 1)) + 1;
```
felsweg-iota: Better move functions, that induce randomness, to some other module.

client/src/procedures.rs
```diff
@@ -0,0 +1,440 @@
...
+        let mut v = Vec::with_capacity(l.into());
+        for _ in 0..l + 1 {
```
felsweg-iota: the increment is not needed here, as you already make sure that `l > 0` will always be true.

client/src/procedures/primitives.rs
```diff
@@ -0,0 +1,1008 @@
...
+impl ProcedureStep for HelperProcedure {
+    fn execute<R: Runner>(self, runner: &mut R, state: &mut State) -> Result<(), anyhow::Error> {
+        match self {
+            HelperProcedure::WriteVault(proc) => proc.execute(runner, state),
+        }
+    }
+}
+
+// TODO: remove notes, add proper docs.
+#[derive(Clone, GuardDebug, Serialize, Deserialize)]
```
felsweg-iota:
```suggestion
#[derive(Clone, GuardDebug, Serialize, Deserialize)]
#[non_exhaustive]
```

To future-proof the APi, better add `#[non_exhaustive]` to the enum, so added variants might be handled with a proper error.

client/src/procedures/primitives.rs
```diff
@@ -0,0 +1,1008 @@
...
+    type Input = ();
+    type Output = ();
+
+    fn generate(self, _: Self::Input) -> Result<Products<Self::Output>, engine::Error> {
+        let mut entropy = [0u8; 32];
+        fill(&mut entropy)?;
+
+        let mnemonic = bip39::wordlist::encode(
+            &entropy,
+            &bip39::wordlist::ENGLISH, // TODO: make this user configurable
```
felsweg-iota: we should somehow tackle this.

client/src/procedures/primitives.rs
```diff
@@ -0,0 +1,1008 @@
...
+            key,
+            output_key: InterimProduct {
+                target: OutputKey::random(),
+                is_temp: true,
+            },
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn dynamic(msg_key: OutputKey, key: Location) -> Self {
```
felsweg-iota: the functions `new` and `dynamic` (with similar intend) appear more than once. Might it be possible to combine them into a trait?

client/src/procedures/primitives.rs
```diff
@@ -0,0 +1,1008 @@
...
+    tag: InputData<Vec<u8>>,
+    nonce: InputData<Vec<u8>>,
+    key: Location,
+    plaintext: InterimProduct<OutputKey>,
+    _marker: PhantomData<T>,
+}
+
+impl<T> AeadDecrypt<T> {
+    pub fn new(
+        key: Location,
```
felsweg-iota, line 910: maybe fewer arguments to the function call? Might there be a possibility to combine them into a parameters struct, that can be reused among individual calls?

client/src/state/secure.rs
```diff
@@ -133,6 +138,119 @@ impl SecureClient {
     }
 }
 
+impl Runner for SecureClient {
```
felsweg-iota, line 141: :+1: for combining relevant function calls into here

felsweg-iota (changes requested):
Looks good to me.

client/src/interface.rs
elenaf9: It would `return anyhow::anyhow!(mailbox_error)`. But I'll add a proper error type for this, instead of using anyhow.

client/src/procedures.rs
elenaf9: But `Into<ProcedureIO>` is not implemented for `Vec<u8>` (and if it was, this `impl From<ProcedureIo> for Vec<u8>` would not be possible because of conflicting implementations). 
Or what exactly do you mean?

client/src/procedures.rs
elenaf9: I agree.

client/src/procedures/primitives.rs
elenaf9: I think we should revisit the `BIP39Generate` Procedure in general. At the moment, the generated mnemonic is only used to derive the seed, an then just thrown away. So technically the language doesn't matter, as it is never exposed to the outside, but at the same time this procedure isn't useful for the user, because they have no way to know the mnemonic. wallet.rs for example doesn't use this procedure, but instead generates the mnemonic outside of stronghold and then uses `Bip39Recover` to create the seed record in stronghold.
I would change the `BIP39Generate` procedure to 1. make the wordlist configurable, and 2. return the mnemonic to the user. What do you think @felsweg-iota?

client/src/procedures/primitives.rs
elenaf9: Generally, I think is a good suggestion and would be desirable to have such a trait. The problem is that the different procedures have different number of input arguments, so adding a trait would require in some cases to wrap the other input parameters  in a new struct, so we can have the signatures:
```
fn new(input: T, others: U) -> Self;
fn dynamic(msg_key: OutputKey, other: U) -> Self
```
with `T` being the input type, and `U`  being other parameters.
I'm afraid that this might make it more complicated for the user when they want to use these functions.

client/src/procedures.rs
felsweg-iota: yes, true that. I have overseen the tuple field inside [`ProcedureIo`]

client/src/procedures/primitives.rs
felsweg-iota: I agree. The wordlist can be passed as parameter but as enum, and the resulting mnemonic should be returned to the user.

client/src/procedures.rs
elenaf9: I don't think there is a fixed convention for this; the traits of `std` are not really consistent on this, e.g.: 

- `Borrow`, `Debug`: Verbs
- `Iterator`: Noun
- `Sized`: Adjective

client/src/interface.rs
tensor-programming: I concur.  A real error type would be useful here.

client/src/procedures/primitives.rs
tensor-programming: I do kind of agree here, it would clean up the code a little to make a decryption input type or something along those lines.

client/src/state/secure.rs
```diff
@@ -74,32 +79,32 @@ impl SecureClient {
     }
 
     /// Resolves a location to a `VaultId` and a `RecordId`
-    pub fn resolve_location<L: AsRef<Location>>(&self, l: L) -> (VaultId, RecordId) {
+    pub fn resolve_location<L: AsRef<Location>>(l: L) -> (VaultId, RecordId) {
```
tensor-programming, line 82: Since `resolve_location`, `derive_vault_id` and `derive_record_id`  no longer need to be methods, might be worth thinking about making them into functions for the Location API or something.   Calling `Self::` on these doesn't make sense when the input/output and logic don't really have anything to do with the `SecureClient`.   Not worth exposing them publicly from the client library but would clean up the code to at least just remove them from the impl block since they are effectively just wrapper functions.

tensor-programming (approved):
Overall looks really nice and well thought out.  Will prove very useful for adding new procs in the future.

client/src/state/secure.rs
felsweg-iota: This will be fixed with another PR.

client/src/actors/secure.rs
```diff
@@ -625,388 +318,10 @@ impl_handler!(
...
-                        vault_id,
-                        record_id,
-                        msg,
-                    },
-                    ctx,
-                )
-            }
-        }
+    fn handle(&mut self, proc: Procedure, _: &mut Self::Context) -> Self::Result {
+        proc.run(self)
```
felsweg-iota, line 325: :heart:

felsweg-iota (approved):
Awesome PR! We now have a (cryptographic) procedures API :partying_face:

PR #269, opened 2021-10-18
Feat: Add differentiated Error types for engine and client

elenaf9:
# Description of change

Remove the single, crate-wide used error types of engine and client. Instead add differentiated error types for the different methods.
Remove `anyhow` Error types in client and bubble up the actual error instead.
Avoid `unwrap`s in the engine where possible.

Use `Vec<u8>` instead of `Location` to specify keys in the Cache.

**Note**: to avoid conflicts with #258, the error types of the procedures are not yet implemented properly. Instead this will be solved in #258 once this PR is merged.

**Note-2**: It already includes PR #270.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Tested via existing tests.

## Change checklist

- [ ] I have performed a self-review of my own code
- [X] New and existing unit tests pass locally with my changes


engine/src/snapshot/kdf.rs
```diff
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// a wrapper around the [`HMAC_SHA256`] function used to derive a hash from a given password.
-pub fn naive_kdf(password: &[u8], salt: &[u8; 32], key: &mut [u8; 32]) -> crate::Result<()> {
+pub fn naive_kdf(password: &[u8], salt: &[u8; 32], key: &mut [u8; 32]) {
```
felsweg-iota, line 5: :+1:  
Good idea to remove returning a `Result` from infallible functions.

client/src/internals/provider.rs
```diff
@@ -24,6 +24,10 @@ impl Provider {
 impl Unpin for Provider {}
 
 impl BoxProvider for Provider {
+    type SealError = crypto::Error;
```
felsweg-iota: Better keep the inner `Error` type generic, but provide a separate enum with `BoxProvider` specific errors. Othwise the inner `Error` type definition would be somewhat redundant. 

But I do get the intention here, though :)

client/src/actors/p2p.rs
```diff
@@ -467,6 +455,53 @@ pub mod messages {
...
+                payload,
+                hint,
+            }
+        }
+    }
+
+    #[derive(DeriveError, Debug, Clone, Serialize, Deserialize)]
+    pub enum RemoteVaultError {
+        #[error("Vault error: {0}")]
+        NotExisting(#[from] VaultDoesNotExist),
```
PhilippGackstatter: Nit: `NotFound` might be a more typical error name?

client/src/actors/snapshot.rs
```diff
@@ -86,25 +86,23 @@ impl Actor for Snapshot {
 
 #[derive(Debug, DeriveError)]
 pub enum SnapshotError {
-    #[error("Could Not Load Snapshot. Try another password")]
-    LoadFailure,
+    // #[error("Could Not Load Snapshot. Try another password")]
+// Engine(#[from] engine::Error),
```
PhilippGackstatter: Leftover comment?

client/src/internals/provider.rs
elenaf9: First of all, I agree that with the current implementation, having 3 associated types that are all `crypto::Error` seems redundant. But this is only the case because `crypto.rs` does not have differentiated error types for each method, but instead only a single, crate-wide error type. Since the engine (and thus the `BoxProvider` trait) may be used independently of our concrete `internals::Provider` implementation in the client, I don't think it should be compromised by this.

> but provide a separate enum with BoxProvider specific errors.

I am not in favor of this idea. I understand that it would reduce the number of associated items and hence clean the trait. But my fundamental goal with this whole PR is specifically that a method will never return an error type / enum where certain variants are never constructed within the method. 
So if in the implementation of `BoxProvider::box_seal` only a type `BoxSealError` can be returned, the user should only ever have to handle the case that `BoxSealError` occurs. If we would now wrap this in an enum  
```
enum ProviderErrors { 
    BoxSealError, 
    RandomnessError  // different error that e.g. BoxProvider::random_buf may return
}
``` 
 we force the user, when calling `BoxProvider::box_seal`, to handle the case of an `RandomnessError` that in reality will never occur.

felsweg-iota: If you use attribute types, that must be specified with a concrete implementation, I currently fail to see where it would make sense to have a separate `OpenError` for example. `BoxProvider` is one unit that declares the interface for safe memory handling. 

> But my fundamental goal with this whole PR is specifically that a method will never return an error type / enum where certain variants are never constructed within the method.
Having a enum type for error types is pretty much common throughout the code, and bundles semantically close errors. 

As an example: `BoxProvider`  can return a `BoxProviderError`, which in turn defines possible variants like `OpenError`, `RandomError`. 

In summary I would argue that an error on type level would suffice, where on function level it is too specific, and would induce a higher coupling to the specific error type without providing much more insight.

elenaf9: > In summary I would argue that an error on type level would suffice, where on function level it is too specific, and would induce a higher coupling to the specific error type without providing much more insight.

For our `client` implementation I agree. But if a user would use the engine directly, I personally favor providing a higher level of granularity for them. Therefore I'd leave it the way that it is currently implemented.
That being said, I also think you have a very good point and I don't really have a strong opinion on this. So please insist in case that you want me to change it.

PhilippGackstatter (commented):
Very nice that you got rid of the `unwrap`s and `line_errors`! You have an interesting approach to error handling. Generally, I like that you make things very explicit in the sense that if a function cannot return an error, it's not part of it's error signature. A potential downside of this approach is on the consumer side. I'll describe our thoughts on this, which might be interesting for you to consider. So far, we had a single enum variant of our error enum (like most libraries), that wrapped the stronghold error type, i.e. `StrongholdError(#[from] iota_stronghold::Error),`. That makes it very convenient to return it from everywhere. This approach no longer works since there are many different `Error<T>` variants now, unless we add an enum variant for each `Error<T>` that currently exists. In our particular case, stronghold is one storage implementation of potentially multiple ones. We are considering adding a `StorageError` type that has fairly generic error kinds, similar to `std::io::ErrorKind`, and map each `Error<T>` variant to that error type. That's still a bit of work, as we need to add implementations for `Error<ReadSnapshotError>` , `Error<WriteSnapshotError>>`  and so on, so we can concretely map a `ReadSnapshotError` to a `StorageErrorKind::ReadError` (or whatever). A blanket implementation wouldn't work, as too much information would be lost. That's our specific thing to deal with, but it might this applies more generally. I think even a library that consumes stronghold directly would want to wrap stronghold errors easily.

It's really cool that your approach is so specific, but composability takes a hit compared to a single global enum. Btw, maybe you haven't seen this yet, and given your approach this could be right down your alley: the [powerset_enum](https://docs.rs/powerset-enum/0.1.0/powerset_enum/index.html) crate. (Doesn't work on stable Rust, though - thanks @olivereanderson for pointing that crate out). Anyhow, I don't think you'll have to change everything to an enum, but I wanted to point out the usability aspects, so you can consider them. If we go with that `StorageError`, we'll match everything to a more generalized error anyway, so what happens underneath wouldn't matter too much for our users.

Are `WriteSnapshotError` and friends actually publicly accessible? `iota_stronghold::WriteSnapshotError` doesn't work, and neither does `iota_stronghold::state::snapshot::WriteSnapshotError`?

client/src/interface.rs
```diff
@@ -150,213 +150,143 @@ impl Stronghold {
...
     /// specified.
-    pub async fn runtime_exec(&self, control_request: Procedure) -> ProcResult {
-        match self.target.send(CallProcedure { proc: control_request }).await {
-            Ok(success) => match success {
-                Ok(result) => result,
-                Err(e) => ProcResult::Error(format!("{}", e)),
-            },
-            Err(e) => ProcResult::Error(format!("{}", e)),
-        }
+    pub async fn runtime_exec(&self, control_request: Procedure) -> Result<ProcResult, Error> {
```
PhilippGackstatter: Is it still appropriate for the `ProcResult` to contain the `Error(String)` variant? One of the main problems with having an error variant there, is that you cannot short-circuit with `?` on that type. Would it be possible to remove that variant and instead move it to a separate error type? Then the signature could be `Result<Result<ProcResult, ProcError>, Error>`.

elenaf9:
@PhilippGackstatter Thanks you for your feedback, it is greatly appreciated! I think your insights from a user perspective are very valuable. I was writing the error types with the assumption that a user would immediately handle/match the error type, but you are right, in practice the error is often wrapped and returned.

> ..., but it might this applies more generally. I think even a library that consumes stronghold directly would want to wrap stronghold errors easily.

I agree. I also did some more reading about rust error-handling and it seems like having a single error type is common practice for libraries, which I wasn't aware of. 
Considering what you wrote, and also the discussion above with @felsweg-iota, I agree that we should refrain from being 100% explicit, and instead favor convenience. I will refactor the types accordingly.
Thanks again for elaboration you point!

---

>  Btw, maybe you haven't seen this yet, and given your approach this could be right down your alley: the powerset_enum crate

I didn't know about that crate, but yes it very much aligned with my error-handling approach. Could be very useful if it would work with stable Rust at some point in the future.

client/src/interface.rs
elenaf9: The Procedure types will be changed in #258 anyway, therefore I did not consider them in this PR. Instead, after the current PR is merged, #258 will merge it and use the new error types

client/src/internals/provider.rs
elenaf9: After thinking about it a bit more I agree with you that in the context of the `BoxProvider` one error type is better.
Especially because in the engine the `BoxProvider` methods aren't used individually, but instead concatenated so that a wrapping error enum would be needed anyway.
https://github.com/iotaledger/stronghold.rs/pull/269/commits/0249d63645e03933f36b50288c9c0a686567012e changes it accordingly.

elenaf9:
So I kept the approach of granular error types instead of single crate-level error enums. 
But I have now flattened and reduced the error types, with some inspiration from the discussion in https://github.com/iotaledger/identity.rs/discussions/460. In particular, the low-level errors in the engine are now abstracted better, and the client doesn't just blindly bubbles them up anymore.

Additionally, I have removed the generic `enum Error<WriteSnapshotError>` pattern that combined shared error types with method-specific ones. Instead, I implement it the more common way by wrapping the `ActorError` and the method-specific error into a new enum:
```rust
#[derive(DeriveError, Debug)]
pub enum WriteSnapshotError {
    #[error("Actor Error: `{0}`")]
    Actors(#[from] ActorError),
    #[error("Write Snapshot Error: `{0}`")]
    Write(#[from] snapshot::WriteError),
}
```
Even though this duplicates a bit of code, I think it is cleaner and easier to understand than the former implementation.

---
The interface now exposes the below Errors, that I believe are relevant for a user.  
Would be great to have some feedback on whether this amount of errors and level of granularity is practical for the user; especially from @PhilippGackstatter .

|Layer | Context | Error Type | 
|- |-|-|
0 | Running Actor System | `MailboxError: {Closed, Timeout}`
1 | Send Message to Actor | `ActorError { MailboxError, TargetNotSpawned }` (Client \|\| NetworkActor)
[*if remote*] 2 | Send Request, Receive Response | `P2PError { ActorError, OutboundFailure {Timeout, DialFailure, ConnectionClosed, UnsupportedProtocols, NotPermitted, Shutdown } } `
[*if can fail*] 3a| Operation on Engine |  see [**Possible Engine Errors**] 
[*if can fail*] 3b| Operation on Swarm | see [**Possible Swarm Errors**]   

**Possible Engine Errors**:
| method | Error Type|
|-|-|
`write_to_vault` | `WriteVaultError::Engine (String)`
`read_snapshot` | `snapshot::ReadError { Io, CorruptedContent, InvalidFile }`
`write_all_to_snapshot` | `snapshot::WriteError { Io, CorruptedData }`

**Possible Swarm Errors**:
| method | Error Type|
|-|-|
`spawn_p2p` | `SpawnNetwork { Io(io::Error) }`
`start_listening` |`ListenErr { Transport { MultiaddrNotSupported, Io }, Shutdown }`
`add_peer` | `DialError { Banned, ConnectionLimit, LocalPeerId, NoAddresses, Aborted, InvalidPeerId, ConnectionIo, Transport, Shutdown }`
`start_relayed_listening` |`ListenRelayErr { DialRelay (DialError), Listen(ListenErr) }`

client/src/interface/errors.rs
```diff
@@ -0,0 +1,162 @@
...
+    use p2p::{DialErr, ListenErr, ListenRelayErr, OutboundFailure};
+    use std::io;
+
+    #[derive(DeriveError, Debug)]
+    pub enum SpawnNetworkError {
+        #[error("Actor Mailbox Error: `{0}`")]
+        ActorMailbox(#[from] MailboxError),
+
+        #[error("Error: Network already running.")]
+        AlreadySpawned,
```
PhilippGackstatter: This is probably a good example of an error that could be considered fatal, if we define "there can only be one running network actor running at any given time" as an invariant. I suppose since the `SpawnNetworkError` type doesn't have a `Fatal` variant, this one could be considered to be that. Maybe think about if it's possible for more fatal errors to be thrown in the future, in which case you may want to name this variant more genericly. 

This also feels like an error that could potentially be avoided by encoding the state of the system in the type system, but I don't know the code base enough to judge properly.

client/src/interface/errors.rs
```diff
@@ -0,0 +1,162 @@
...
+use actix::MailboxError;
+use std::fmt::Debug;
+use thiserror::Error as DeriveError;
+
+#[derive(DeriveError, Debug)]
+pub enum ActorError {
+    #[error("Error sending message to Actor: `{0}`")]
+    Mailbox(#[from] MailboxError),
+    #[error("Target Actor has not been spawned or was killed.")]
+    TargetNotSpawned,
```
PhilippGackstatter: Nit: Maybe `TargetNotFound` would be clearer. It contains the two cases "not spawned" and "killed" better? I would still leave the error message as is.

client/src/interface/errors.rs
```diff
@@ -0,0 +1,162 @@
...
+        #[error("Local Actor Error: `{0}`")]
+        LocalActors(#[from] ActorError),
+        #[error("Listen Relay Error: `{0}`")]
+        ListenRelay(#[from] ListenRelayErr),
+    }
+
+    #[derive(DeriveError, Debug)]
+    pub enum P2PError {
+        #[error("Local Actor Error: `{0}`")]
+        LocalActors(#[from] ActorError),
```
PhilippGackstatter: Nit: Would `LocalActor` be better?

engine/src/vault/view.rs
```diff
@@ -9,11 +9,39 @@ use crate::vault::{
...
+    Procedure(TProcErr),
+}
+
+#[derive(DeriveError, Debug)]
+pub enum RecordError<TProvErr: Debug> {
+    #[error("Provider Error: `{0:?}`")]
+    Provider(TProvErr),
+
+    #[error("Found Invalid Transaction")]
+    CorruptedContent,
```
PhilippGackstatter: Since this is basically a fatal error, it would be helpful for it to contain more information. Alternatively, since fatal errors require human intervention, it also seems okay to log additional information.

engine/src/vault/view.rs
```diff
@@ -9,11 +9,39 @@ use crate::vault::{
...
+
+    #[error("Invalid Key provided")]
```
PhilippGackstatter: Error messages should be lowercase (we're also doing this wrong :sweat_smile:). I'm fairly sure this is mentioned somewhere in error handling best practices, but can't find it right now. Instead, quoting _Rust for Rustaceans_:

> In general, your implementation of Display should give a one-line description of what went wrong that can easily be folded into other error messages. The display format should be **lowercase** and without trailing punctuation so that it fits nicely into other, larger error reports.

PhilippGackstatter (commented):
Overall, I think this looks very nice. This should fit nicely into our new error handling approach, but we'll be able to provide more feedback when actually upgrading to the later version with these error types, which is currently not possible due to time constraints. I hope that's okay. I think it's still acceptable to make some small change to the error types during the release candidate phase.

client/src/interface/errors.rs
elenaf9: I find these types of errors a bit hard to classify. I wouldn't really consider it a fatal error, since a imo a fatal error is an unexpected failure from which you can not recover.  The `AlreadySpawned` error is an expected error if you don't follow the invariant that you have mentioned above. Ideally, the code that is responsible for spawning the network (could be your code, or the code of the end-application) should take care that this error can never happen, so that it can be treated as `unreachable`. 
Apart from that, depending on the intention of the caller, the error may not even matter in some cases, e.g. when you use it as an "spawn a new p2p-actor if there isn't one already" (though one can argue that using it like that isn't very clean - but still, it's allowed).

---
Maybe that's even something to also add to your list in the discussion https://github.com/iotaledger/identity.rs/discussions/460 on how to classify errors: **"Expected vs Unexpected"** errors. It would also apply for e.g. `ActorError::TargetNotFound` or `OutboundFailure::NotPermitted`, which are also errors where, depending on the implementation, you known for sure they can never occur.

> This also feels like an error that could potentially be avoided by encoding the state of the system in the type system, but I don't know the code base enough to judge properly.

That would be desirable yes, but not sure how to do that. Do you have an example from e.g. the identity code base where something like this is done?

PhilippGackstatter: Depending on whether you define your invariant one way or another, it's expected (local) or unexpected (fatal), so I think it's another way to look at the same consideration, which is still useful!

But it's fair to say that this is more of a case where the user needs to define whether this is an invariant, and unwrap or map it to unreachable. 

> Do you have an example from e.g. the identity code base where something like this is done?

The basic idea is to call `spawn` which returns a, say, `StrongholdP2p` type. On that type, you can not call `spawn` again, and so you don't have this error. This discourages accidentally calling spawn twice, but if it happens anyway, this then returns a lower-level error such as `AdressInUse`.

I've tried to do this with the identity actor, where creating an `Actor` means you also start to [spawn the listener](https://github.com/iotaledger/identity.rs/blob/feat/identity-actor/identity-actor/src/actor/actor.rs#L95) in the background. It guarantees any actor type you have is always listening, and you can't forget to start listening or accidentally try to spawn a listener twice. (You can still forget to listen on an address, though, which is a different topic).

The _type system guidance_ section in chapter 3 of _Rust for Rustaceans_ talks about another example for state machines with more states, which may also be interesting to look at.

elenaf9:
_Sorry for yet another change._

Because the `ActorError` can occur on every operation in the interface, we added  a nested Result type alias `StrongholdResult<T> = Result<T, ActorError>`, with `T` being the actual return value of this method. E.g. 
```rust
fn read_from_store(..) -> StrongholdResult<Option<Vec<u8>>> {..}

fn delete_from_store(..) -> StrongholdResult<()> {..}

fn read_snapshot(..) -> StrongholdResult<Result<(), ReadError>> {..}
```
The  error enums that I mentioned above (to wrap the `ActorError` with method-specific errors, e.g. `WriteSnapshotError`) were removed in favor of this new type.
Apart from avoiding redundant code, this design has the benefit that the user can easily return the `ActorError` with the try operator `?` and then can handle the method error directly.

---
Unless there are any objections, this latest design will now be considered final.

client/src/lib.rs
```diff
@@ -39,29 +37,34 @@ mod tests;
...
+    };
+}
 
-pub use engine::vault::RecordHint;
+pub use actix::MailboxError;
 
 /// TODO: Should be replaced with proper errors.
+#[cfg(test)]
 #[macro_export]
 macro_rules! line_error {
```
felsweg-iota: With the new error types we may consider removing this macro.

felsweg-iota (approved):
The final changes to the error types look good to me. The result is a good compromise between (complex) module level error types, and crate level error types.

PR #270, opened 2021-10-21
Use Registry for all actors

elenaf9:
# Description of change

Also maintain network actor and current client target in the registry.
Make current client-target optional, in case that it is killed before switching to another target.
Make registry a normal actor instead of a system-service.

**Note:** Proper erorr handling will be included in #269.

## Type of change

Tested via existing tests

## Change checklist

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/client/src/actors/registry.rs
+++ b/client/src/actors/registry.rs
@@ -14,7 +14,9 @@ use engine::vault::ClientId;
 use std::collections::HashMap;
 use thiserror::Error as ErrorType;
 
-use crate::{actors::SecureClient, state::snapshot::Snapshot};
+#[cfg(feature = "p2p")]
+use super::p2p::NetworkActor;
+use crate::{actors::SecureClient, internals, state::snapshot::Snapshot};
 
 #[derive(Debug, ErrorType)]
 pub enum RegistryError {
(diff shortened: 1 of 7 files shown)


client/src/interface.rs
```diff
@@ -68,17 +67,15 @@ impl Stronghold {
...
         };
 
-        Ok(Self {
-            registry,
-            target,
-            #[cfg(feature = "p2p")]
-            network_actor: None,
-        })
+        let mut sh = Self { registry };
+        sh.switch_client(client_id).await;
```
felsweg-iota: On line 70 you already spawn the client, better use the returned address directly :)

client/src/interface.rs
```diff
@@ -89,56 +86,30 @@ impl Stronghold {
...
-                                ))
-                            }
-                        }
-                    }
-
-                    self.target = client
-                }
-                None => return StatusMessage::Error("Could not find actor with provided client path".into()),
-            }
+        match self.switch_client(client_id).await {
```
felsweg-iota, line 99: Nice!

client/src/actors/registry.rs
```diff
@@ -72,14 +82,54 @@ pub mod messages {
     impl Message for GetAllClients {
         type Result = Vec<(ClientId, Addr<SecureClient>)>;
     }
+
+    #[cfg(feature = "p2p")]
```
felsweg-iota: you can avoid having multiple messages types annotated with a feature gate by bundling them into a separate module and re-import it on super module level.

felsweg-iota (approved):
Moving the current actor target to the registry was definitely a good idea, also having a registry per arbiter will increase isolation. Good work!

PR #277, opened 2021-11-03
test(client): Restrict length of random key

elenaf9:
# Description of change

Fix test in client: for the key a Vec with a random length was used. 
Since this length can be up to `usize::Max`, it tried to allocate more memory than available, hence causing:
```
...
memory allocation of 11744374998907113874 bytes failed
error: test failed, to rerun pass '-p iota_stronghold --lib'

Caused by:
  process didn't exit successfully: `/home/runner/work/stronghold.rs/stronghold.rs/target/debug/deps/iota_stronghold-f2cb821b8be1a728` (signal: 6, SIGABRT: process abort signal)
```
With this PR, the key will have a max length of 1024B.

Amend Changelog of #269.

Wrap the String message of fatal vault errors in newtype `FatalEngineError`.

## Type of change

### DIFF
--- a/.changes/refactor-error-types.md
+++ b/.changes/refactor-error-types.md
@@ -1,0 +1,13 @@
+---
+"iota-stronghold": minor
+"stronghold-engine": minor
+---
+
+
+[[PR 269](https://github.com/iotaledger/stronghold.rs/pull/269)]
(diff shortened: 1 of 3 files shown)


felsweg-iota (approved):
Looks good to me!

Issue #281, opened 2021-11-05
Persist state of Stronghold-P2p 


elenaf9:
## Description

Allow the export of the Stronghold P2p behaviour state (Firewall configuration, known addresses of remote peers, etc.) when shutting down the local peer. Enable the user to import a former state on init when starting a new local peer.
Within Stronghold, the interface should provide methods to write the exported state into the Stronghold store to persist it, and load it back from the store when spawning a new `NetworkActor`. 
Specifically for the underlying keypair that is used to authenticate the network, Stronghold should support re-use of the same keypair by storing it in the vault, and provide a Procedure to derive a new Noise-Keypair from it for each new session.

## Motivation

In most cases, the configuration set in Stronghold-P2p should not only be valid for a single session, but instead is a general setting that should apply each time the network is running. The user should not have to manually set every rule and information again each time they spawn a local peer, especially if these are dynamic configs that were set by the end-user.
In case of the authentication-keypair, the Peer-Id is derived from this keypair. Since the Peer-Id serves as unique identifier of our local Peer/ Stronghold and is used by remote peers to set firewall rules and store address information, it should stay the same over the course of multiple sessions. 

## Requirements

1. Allow export of current state/ config of Stronghold-P2p in a serializable/ deseriaizable format
2. Enable import of a former state into Stronghold-P2p 
3. Stronghold client should enable persisting the state in the Stronghold store
4. Stronghold should provide Procedures to generate a new p2p authentication keypair in the vault, and derive Peer-Id and noise-keypair from it

## Are you planning to do it yourself in a pull request?

Yes

PR #285, opened 2021-11-10
feat(client): persist p2p keypair in vault

elenaf9:
# Description of change

Allow generating and storing the ed25519 [identity-keypair](https://docs.rs/libp2p/0.40.0/libp2p/identity/enum.Keypair.html) for p2p in the vault. The `PeerId` of the local Stronghold is derived from this keypair, hence it is desirable to persist it. Allow deriving a new authentic [noise-keypair](https://docs.rs/libp2p/0.40.0/libp2p/noise/struct.AuthenticKeypair.html) from the keypair, that is used for the authentication and noise-encryption on the transport layer.

First step towards #281.

## Type of change

### DIFF
--- a/.changes/persist-p2p-keypair.md
+++ b/.changes/persist-p2p-keypair.md
@@ -1,0 +1,7 @@
+---
+"iota-stronghold": patch
+---
+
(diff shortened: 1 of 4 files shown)


felsweg-iota (approved):
Looks good to me!

PR #290, opened 2021-11-16
Persist config and keypair of stronghold-p2p in client

### LINKED ISSUE #281: Persist state of Stronghold-P2p

elenaf9:
# Description of change

Persist the state of stronghold-p2p in the client by serializing the `NetworkConfig` and writing it to the store.
Allow loading stored states into the network actor on init.
Allow reuse of same `Keypair` (for authentication and encryption of the p2p transport layer) that is stored in the vault, to keep the same `PeerId` over multiple sessions.

**Note**: Firewall rules of type `Rule::Restricted` can not be serialized and hence are skipped when serializing the config.
This is not optimal, but at the current point there's no real solution for this.

Fixes issue #281 .

## Type of change

- [x] Documentation Fix

Added new test `interface_tests::test_p2p_config`.

## Change checklist

### DIFF
--- a/client/examples/p2p/main.rs
+++ b/client/examples/p2p/main.rs
@@ -30,7 +30,6 @@ use arguments::*;
 pub use clap::Parser;
 use iota_stronghold::p2p::{Multiaddr, NetworkConfig, SwarmInfo};
 pub use iota_stronghold::Stronghold;
-use p2p::firewall::Rule;
 pub use std::error::Error;
 
 /// Returns a list of all available peers
@@ -52,7 +51,7 @@ pub async fn list_peers_command(stronghold: &mut iota_stronghold::Stronghold) ->
 
 /// Displays the swarm info of this stronghold instance
 pub async fn show_swarm_info_command(stronghold: &mut iota_stronghold::Stronghold) -> Result<(), Box<dyn Error>> {
-    stronghold.spawn_p2p(Rule::AllowAll, NetworkConfig::default()).await?;
+    stronghold.spawn_p2p(NetworkConfig::default()).await?;
 
     let SwarmInfo {
         local_peer_id,
@@ -77,7 +76,7 @@ pub async fn start_listening_command(
     let multiaddress: Multiaddr = multiaddr.parse()?;
 
     // spawn network actor
-    let network = stronghold.spawn_p2p(Rule::AllowAll, NetworkConfig::default()).await;
+    let network = stronghold.spawn_p2p(NetworkConfig::default()).await;
     println!("Network actor spawned: {:?}", network);
 
     // start listening
--- a/client/examples/p2p/main.rs
+++ b/client/examples/p2p/main.rs
@@ -51,7 +51,7 @@ pub async fn list_peers_command(stronghold: &mut iota_stronghold::Stronghold) ->
 
 /// Displays the swarm info of this stronghold instance
 pub async fn show_swarm_info_command(stronghold: &mut iota_stronghold::Stronghold) -> Result<(), Box<dyn Error>> {
-    stronghold.spawn_p2p(NetworkConfig::default()).await?;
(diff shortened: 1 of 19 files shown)


client/examples/p2p/main.rs
```diff
@@ -77,7 +76,7 @@ pub async fn start_listening_command(
     let multiaddress: Multiaddr = multiaddr.parse()?;
 
     // spawn network actor
-    let network = stronghold.spawn_p2p(Rule::AllowAll, NetworkConfig::default()).await;
+    let network = stronghold.spawn_p2p(NetworkConfig::default(), None).await;
     println!("Network actor spawned: {:?}", network);
```
felsweg-iota, line 80: Better remove print statements :)

p2p/src/serde.rs
```diff
@@ -0,0 +1,176 @@
...
+use serde::{Deserialize, Serialize};
+use smallvec::SmallVec;
+use std::{
+    collections::HashMap,
+    convert::{TryFrom, TryInto},
+    marker::PhantomData,
+};
+
+#[derive(Serialize, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct SerdePeerId(Vec<u8>);
```
felsweg-iota, line 19: Discussion: 
You use wrapper types to simplify de/serialization of inner types like `PeerId`, `AddressInfo` etc. Do you think it might reasonably reduce the amount of wrapper types, but using a `DeSerializeFirewallConfig` enum type? The variants could capture all necessary data enabling de/serialization with `serde`; providing a `From` implementation would bind to all relevant types.

felsweg-iota (approved):
Looks good to me! Definitely a useful addition. :heart:

p2p/src/serde.rs
elenaf9: If `PeerId` would implement `Serialize` and `Deserialize`, we could remove most of these workaround types. I have opened an Issue at libp2p https://github.com/libp2p/rust-libp2p/issues/2341 and will work on a PR to maybe solve this.

elenaf9: I would merge this for now how it is, and resolve this once the changes will (hopefully) be added in libp2p. Okay @felsweg-iota?

client/examples/p2p/main.rs
elenaf9: Don't we want to have some print statements in examples?

p2p/src/serde.rs
felsweg-iota: yes please

PR #293, opened 2021-11-23
fix(p2p): fix EventChannel, add tests

elenaf9:
# Description of change

Refactor `Sink` and `Stream` implementation for `EventChannel`.
The `Sink` implementation allows to write a message to the channel (blocking or non-blocking).
In case of an non-blocking `Sink`, the `Stream` implementation has to be continuously polled. The polling drives the sending of messages once the inner channel has the capacity to receive it (in the `SwarmTask`, this polling is done in the central loop in `Stream::run`).
The `ChannelSinkConfig` allows to configure the channels behaviour:
- in case of ` ChannelSinkConfig::Block`: 
  - `Sink` implementation delegates every call to the inner `mpsc::Sender`
  - if the inner channel has capacity for the message: the message is send and  `EventChannel::send` returns
  - if the inner channel is full: `EventChannel::send` blocks until the message can be send
  - `Stream` implementation is not relevant.
-  in case of `ChannelSinkConfig::DropLatest | ChannelSinkConfig::BufferLatest`:
   - if the inner channel has capacity for the message:
      - start sending the messages
      - polling  `<EventChannel as Stream>` flushes the message 
   - if the inner channel is full: 
     -  if `ChannelSinkConfig::BufferLatest`
        - Write message to buffer
        - polling `<EventChannel as Stream>` writes from buffer to inner channel  once there is capacity
     -  if `ChannelSinkConfig::DropLatest`: drop message

## Type of change

Added `event_channel::test` module to test every `ChannelSinkConfig`.

## Change checklist


p2p/src/interface/event_channel.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// New events will be dropped if the channel is full.
+    DropLatest,
+    /// In case that the channel is full, store new events in a ring-buffer. If the configured capacity is reached,
+    /// older events will be dropped in favor of newer ones. Send the latest events sequentially in FIFO order when
+    /// the channel has free capacity.
+    BufferLatest,
+}
+
+/// Wrapper of a [`mpsc::channel`][futures::channel::mpsc::channel] for sending events.
+/// The [`EventChannel`] cen be configure its behaviour e capacity of the channel is reached.
```
felsweg-iota: sorry, I do not understand this comment. Is this meant:
"The [`EventChannel`] can configure its behavior until the capacity of the channel is reached." ?

p2p/src/interface/event_channel.rs
```diff
@@ -0,0 +1,299 @@
...
+    // Wether the `Sink` implementation of the inner channel should be used, without a buffer.
+    //
+    // This results in the `SwarmTask` blocking until `<mpsc::Sender as Sink>::send` resolves.
+    use_inner: bool,
+    // Waker from `<EventChannel as Stream>::poll_next` that is notified if a new event was added to the buffer.
+    waker: Option<Waker>,
+}
+
+impl<T> EventChannel<T> {
+    pub fn new(capacity: usize, config: ChannelSinkConfig) -> (Self, mpsc::Receiver<T>) {
```
felsweg-iota, line 52: Since an `EventChannel<T>` can only have the values of `ChannelSinkConfig` this implementation is Ok, but I would suggest to leave room for extension: Reduce the responsibility of `new` to one, create new functions like `new_blocking(..)`, `new_drop_latest(...)`, etc...

felsweg-iota (changes requested):
May I also ask you to describe the bug in a few sentences, that has been fixed? Thank you :)

p2p/src/interface/event_channel.rs
elenaf9: I don't understand it myself either. Must have been my cat walking over the keyboard 😄.

elenaf9:
> May I also ask you to describe the bug in a few sentences, that has been fixed? Thank you :)

The bug was caused by a fundamental misunderstanding on my side of what `Sink::poll_flush` does. In the old implementation I assumed that  "flushing" should do something like clear all pending requests, hence I actually just cleared the buffer and then called `inner.poll_flush`. In reality `poll_flush` is responsible for doing the actual writing of the request and in case of `mpsc::Sender` it blocks the `send` operation until the channel has capacity to received the request. Now with this PR it is implemented that `poll_flush` only blocks in case of `ChannelSinkConfig::Block`, otherwise it immediately returns and the send-request-once-channel-has-capacity part is done in the `Stream` implementation

p2p/src/interface/event_channel.rs
elenaf9: I personally favor having a single new method with an enum, over having a separate `new` method for each variant. Especially since the rest of the parameters (in this case only `capacity`) is the same for all variants, hence all 3 methods would have the same signature.

felsweg-iota (approved):
:heart:

PR #294, opened 2021-11-25
feat(client): allow write snapshot with stored key

### LINKED ISSUE #279: Store encryption in vault for continuous writing to snapshot on disk
## Description

Writing a snapshot to disk requires a key. Holding this key in memory outside Stronghold is insecure. Provide the means to keep a key inside the vault to enable  writing (to) a snapshot without further providing a key. 

## Motivation

This would enable partial automation of persisting snapshots without keeping the key inside an unsafe memory location. 

## Requirements

1.  extend interface to store a snapshot key, and store the location during the Stronghold session. Make sure that the key itself won't get written into the snapshot itself.

## Open questions (optional)

none. 

Yes.

elenaf9:
# Description of change

Provide method for writing the snapshot with a key that is stored in the vault.
Extract the key from the client in which it is stored in and use it as encryption key to write all data to the snapshot.

Fixes issue #279.

## Type of change

Added new test `basic_tests::write_snapshot_with_stored_key`.

## Change checklist

### DIFF
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -22,12 +22,12 @@ use crate::{
     procedures::{CollectedOutput, Procedure, ProcedureError},
     state::{
         secure::SecureClient,
-        snapshot::{ReadError, WriteError},
+        snapshot::{ReadError, WriteError, WriteWithKeyError},
     },
     utils::{LoadFromPath, StrongholdFlags, VaultFlags},
     Location,
 };
-use engine::vault::{ChainId, ClientId, RecordHint, RecordId, VaultId};
+use engine::vault::{ClientId, RecordHint, RecordId};
 #[cfg(feature = "p2p")]
 use p2p::{identity::Keypair, DialErr, InitKeypair, ListenErr, ListenRelayErr, OutboundFailure, RelayNotSupported};
 
@@ -130,20 +130,6 @@ impl From<String> for FatalEngineError {
     fn from(e: String) -> Self {
         FatalEngineError(e)
     }
-}
-
-#[derive(DeriveError, Debug)]
-pub enum WriteSnapshotWithKeyError {
-    #[error("write snapshot failed: {0}")]
-    WriteSnapshot(#[from] WriteError),
-    #[error("fatal engine error: {0}")]
-    Engine(#[from] FatalEngineError),
-    #[error("vault not found: {0}")]
-    VaultNotFound(VaultId),
-    #[error("record not found: {0:?}")]
-    RecordNotFound(ChainId),
-    #[error("invalid key: {0}")]
-    InvalidKey(String),
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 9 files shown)


client/src/interface.rs
```diff
@@ -387,6 +401,58 @@ impl Stronghold {
         Ok(res)
     }
 
+    /// [`Stronghold::write_all_to_snapshot`] using a key for encryption that is stored in at the specified client and
+    /// location.
+    pub async fn write_snapshot_with_stored_key(
+        &mut self,
+        key: Location,
```
PhilippGackstatter, line 416:
```suggestion
    /// [`Stronghold::write_all_to_snapshot`] using a key for encryption that is stored at `key_location`
    /// in the specified client.
    pub async fn write_snapshot_with_stored_key(
        &mut self,
        key_location: Location,
```

PhilippGackstatter (approved):
Looks great from the user perspective! Would simplify using stronghold quite a bit.

tensor-programming (approved):

tensor-programming:
Looks good from what I can see.

tensor-programming:
Going back through this PR, there are a couple thoughts that come to mind.  

Writing the password into the snapshot itself seems like a bad idea any way you slice it.  That makes me wonder if there is some way to distance this password record from the rest of the internal stronghold state.  

Tangential to this PR: I also wonder if there is a way that we can track the change in state between writes.  The real way that someone could mess with a stronghold is by changing the data inside or using the data through procedures, which means that reading or writing to the snapshot is not inherently insecure (assuming the memory guards and encryption holds).  If the state hasn't changed then the write doesn't really effect the snapshot in the sense that the system is effectively just rewriting the data into the snapshot. And even in the case of firefly, the main usage of the system is to derive verify and sign data while creating new addresses for the wallet.  There are instances where someone opens the snapshot and the stronghold and does nothing with it that changes the state inside of it.  If we could track the state, we could throw out the password record and ask for the password again upon write because the state changed.  Or we could have the password work on a time period where the password in this record has a kind of TTL which drops the record after that time period requiring the user to reauthenticate.

client/src/interface.rs
```diff
@@ -409,6 +409,37 @@ impl Stronghold {
...
+        path: Option<PathBuf>,
+    ) -> StrongholdResult<Result<(), WriteWithKeyError>> {
+        let client_id = ClientId::load_from_path(&client_path, &client_path);
+
+        let clients: Vec<(ClientId, Addr<SecureClient>)> = self.registry.send(GetAllClients).await?;
+
+        let snapshot = self.registry.send(GetSnapshot {}).await?;
+
+        for (id, client) in clients {
+            let data = client.send(GetData {}).await?;
```
felsweg-iota, line 428: This part of exporting client's data is crucial, as it contains the key location. At this explicit part I would suggest to extract  the key from client's state, write the snapshot without the decryption key inside it, and reinsert it again.

felsweg-iota (changes requested):
I agree with @tensor-programming on this part, that storing the actual en-/decryption key inside the snapshot might not be a good idea. A better approach could be to let the key "live" on a per-session basis, removing it on I/O operations on the snapshot. The session key could live on a secure self-evicting cache, possibly an extension to the vault. Access to the key could be restraint by time or number of access. This would ease the use of writing snapshots more often or in automated way, while keeping the actual snapshot password secure.

felsweg-iota:
> If we could track the state, we could throw out the password record and ask for the password again upon write because the state changed.

This would resemble some kind of log of snapshot `operations` that can be `committed` and bundled up in a `transaction`. The `operations` would resemble a diffed state.

elenaf9:
Closing this PR, because this will be solved in the upcoming implementation of the RFC described in #308.

PR #292, opened 2021-11-17
Feature: Policy engine with state synchronization

felsweg-iota:
# Description of change

This PR brings in a lightweight policy engine to configure dynamic access control. This is especially useful, when requesting state synchronization for remote peers, when the requesting party should only be allowed to access certain locations inside the serving party. 

## Links to any relevant issues

This PR solves the issues [Policy engine](https://github.com/iotaledger/stronghold.rs/issues/280) and [State Synchronization Across Multiple Stronghold Instances](State Synchronization Across Multiple Stronghold Instances). 

## Type of change

- [x] Enhancement (a non-breaking change which adds functionality)

## How the change has been tested

integration and unit tests inside `policy` and `client` crate. 

## Change checklist

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #296, opened 2021-11-28
[Proposal] Improve stronghold interface

PhilippGackstatter:
# Description of change

The identity.rs and wallet.rs crates wrap stronghold in very similar ways. This led us to align on what our needs are and it appears that we could use the same wrapper crate, if it existed. I looked into creating that crate based on what we already have in identity.rs. During that process, I noticed some shortcomings in the current stronghold interface, that I would love to see fixed. Rather than fixing it by wrapping stronghold in a new crate, I tried to improve the interface itself. These are the problems I was trying to solve:

1. Because only one actor can be the target and it needs to be switched through the interface, there is an inherent bottleneck. Users who want to access stronghold with multiple clients concurrently need to wrap the `Stronghold` into a mutex, to ensure that switching the target _and_ a subsequent operation is one atomic operation. If mutually exclusive access is not guaranteed, another concurrent caller may switch the target to some other actor; a race condition. This target mechanism seems unnecessary, given that the underlying actor system already handles synchronization, and hinders usability and potentially performance.
2. Everything goes through the `Stronghold` type. Stronghold has a very powerful architecture, multiple levels deep, but all of it passes through a flat interface. That means the interface misses out on the benefits of the strong type system that Rust offers. For example, actors need to be managed explicitly, through spawn and kill operations. Method names tend to be long and are sometimes ambiguous, i.e. is that method operating on the store or vault? Due to the target mechanism, it's not immediately obvious what client we operate on, which makes stronghold code harder to read and learn.
3. Users need to manage passwords. Both identity.rs and wallet.rs have stronghold wrapper modules that take care of storing and clearing passwords, in order to read and write snapshots. These passwords are stored in unguarded memory. With #294 taking care of writing snapshots, the only place left were passwords are needed repeatedly is in `read_snapshot`. Stronghold should make it easy to do the right thing, and it would ideally only require the password from the user on startup, in order to decrypt the snapshot, and then never again.

To address these shortcomings, I refactored the interface, while keeping practically the same functionality. This is an example:

```rust
#[actix::test]
async fn stronghold_interface_example() -> Result<(), Box<dyn std::error::Error>> {
    let vault_path = b"vault".to_vec();
    let client_path = b"client";
    let key_data = [0xff; 32].to_vec();
    let file = SnapshotFile::named("testfile");

    let record_location = VaultLocation::counter(0);

    let store_loc = bytestring(4096);

    let mut snapshot: Snapshot = Snapshot::new(file.clone());
    let client: Client = snapshot.client(client_path).await?;
    let vault: Vault = client.vault(vault_path.clone());
    let store: Store = client.store();

    vault
        .write(
            record_location.clone(),
            b"test".to_vec(),
            RecordHint::new(b"first hint").unwrap(),
        )
        .await??;

    store.write(store_loc.clone(), b"test".to_vec(), None).await?;

    snapshot.write(&key_data).await??;

    std::mem::drop(client);
    std::mem::drop(snapshot);

    let mut snapshot = Snapshot::new(file.clone());
    snapshot.read(&key_data).await??;

    let client: Client = snapshot.client(client_path).await?;
    client.restore_state().await?;

    let store_data = client.store().read(store_loc.clone()).await?.unwrap();
    let vault_data = client
        .vault(vault_path)
        .read_secret(record_location.clone())
        .await?
        .unwrap();

    assert_eq!(store_data, b"test");
    assert_eq!(vault_data, b"test");

    Ok(())
}
```

This addresses the above-mentioned issues:

1. This interface exposes actors more directly via the `Client` type. One client is backed by one actor. `Stronghold::client` can be called multiple times with the same `client_path` to get another `Client` that is backed by the same actor, and the client is cloneable. This way, multiple clients can be created and used from different tasks and threads, granting the user more direct access to the power of the actor system. This design also gets rid of (most of) the synchronization that callers have to do, because the actor system itself already handles that. That improves performance (fewer locks needed) and, perhaps more importantly, usability.
2. Calling `Stronghold::client` conceptually does a `get_or_init` operation for the actor, in order to avoid having to explicitly spawn an actor. Moreover, a client represents a running actor, so `TargetNotFound` errors become unnecessary (however, they are still in the code for short-on-time reasons). Additionally, the interface gains a `Vault` and `Store` type, which can be created on a client. In this interface, it's unambiguous whether you operate on a store or a vault, and the interface is easier to read and reason about. For example, it's clear which client a store or vault belongs to, while in the current interface one has to find the last call to `switch_actor_target` to find out.
3. In the current approach, the `ReadFromSnapshot` actor message does two things: find the actor state of the passed `client_id` in the in-memory snapshot state and return it, or read the snapshot state from file into memory and then return the actor state. These two concerns can be separated into separate actor messages. In the refactored interface, `Stronghold::read_snapshot` reads the file from disk, requiring the key for decryption. Then, a client can call `Client::restore_state` to copy the data from the snapshot into the actor, but without requiring the password. Now the password is only required until the snapshot has been read. Any subsequent call to `restore_state` uses what is already in-memory. If someone actually needs to read actor states from different snapshots, they have to call `Stronghold::read_snapshot` and `Client::restore_state` multiple times, which is roughly the same amount of effort as the current interface requires - and that seems fine. Crucially, this approach makes life much simpler for those that only use a single snapshot (like identity.rs and (I believe) wallet.rs).

So those are all the nice parts about the refactor. What follows are the challenging parts.

## Actor state

If actors are referenced from `Client`s, then once all clients have gone away, we would like to also remove the actor from the `Registry`, simply to clean up and reduce the number of `GetData` and `FillSnapshot` roundtrips we need to make when writing a snapshot. I went through several ideas before arriving at something that was sort of acceptable.
The registry ideally only holds active actors. In the current interface, it's the user's responsibility to spawn and kill actors when appropriate. With the introduction of the `Client`, we cannot explicitly kill an actor by implementing something like `fn kill(self)` on `Client`, because one can get multiple `Client`s backed by the same actor from stronghold using `Stronghold::client`. So it's always possible to have multiple clients that use the same underlying actor. If a `Client` represents a running actor and we have `let client2 = client1.clone()`, then `client1.kill()` would leave a `client2` in an undefined state. Here are some ideas to fix that, and the problems with those...

1. However we detect it, when the last client (internally holding an `Addr`) to an actor is cleaned up, we would like to write the actor's data to the `SnapshotState`. Then, `write_snapshot` only needs to take the data and write it to disk. By only storing `WeakAddr`s in the registry itself, all clients can hold strong references (`Addr`) to actors. (A weak address doesn't prevent an actor from being dropped, unless one or more strong `Addr`s are still held.) Even if we figure out when an appropriate time is to move the data from the actor to the snapshot, running it is very tricky. Due to the lack of an `AsyncDrop`, copying the data can only be done in the sync `Drop` impl. One idea is to send a single `do_send` (sync) message to the `Snapshot` actor, which itself gets the data from the actor and fills the snapshot. That also involves async operations in the sync `handle`, which I tried running with `ctx.wait` but ended up with a deadlock. Similarly, the same should be possible by implementing `Actor::stopping` and transfer the data there. Since that would have access to the actor itself, the data could be copied and sent off with `do_send`, however no errors could be handled. Considering that stronghold manages private keys, a solution needs to be more resilient. So even when `AsyncDrop` lands, the lack of error handling or being able to return them, remains a problem.
2. Another option is to use strong references in the registry and make us of weak addresses to figure out what actors are still alive. During `write_snapshot`, all addresses in the registry are downgraded to `WeakAddr`s, and then attempted to be upgraded to strong addresses again. If the upgrade fails, we know that no other references to the actor exist and we can remove the client from the registry. Because the registry stores strong addresses, even if all `Client`s for an actor have gone away, the data isn't lost. `write_snapshot` writes the data to disk, and only then allows the unreferenced actors to be deallocated. The cool part is that the user doesn't need to explicitly kill an actor, and can't forget to kill it, it is automatically dropped as soon as it's no longer needed. This is basically the lazy version of option 1.
   This solution technically works, but it requires the `Stronghold` to not be `Clone` and `write_snapshot` to take `&mut self`. From my understanding, the down- and upgrading cannot be done on the actor thread, because an actor handler is an atomic operation from actix' point of view, and thus the actix runner will never observe that the last reference to an actor has been removed; preventing deallocation. This means the down- and upgrading needs to happen outside the actor thread, i.e. within `Stronghold::write_snapshot`, but in one atomic operation. That's why `&mut self` and the absence of `Clone` are required.
3. Live with the downsides of option 2, i.e. the non-cloneable `Stronghold`, and let users deal with it. They will have to wrap it in an `Arc<Mutex<Stronghold>>` in order to use it concurrently. Single-threaded users on the other hand don't have to deal with it specially. Since one of the primary reasons for doing all this was to make stronghold easier to use from concurrent contexts, I'm not super happy about this. I still think it would be an overall improvement, though, because once a `Client` has been obtained, that client can be used from any task or thread. (This is the currently implemented solution.)
4. Option 2, but add an internal lock to `Stronghold` itself and take care of the problem for the users. We need to lock whenever we do an operation that involves clients, so in `Stronghold::client` and `Stronghold::write_snapshot`. 
5. Don't bother dealing with all that automatic handling of copying actor state to snapshots and let users deal with it explicitly. That is, they have to remember to call `client.save_state()` when they want state to be persisted. This is simple to implement for the library, but seems like a footgun and makes stronghold harder to use correctly.

## Procedures

In the example, there is a new type called `VaultLocation`. It is basically a `Location` but without the `vault_path`, because that is implicit when we call `Vault::write`, for example. The procedures all use `Location`, and it's much harder to replace those occurrences. It would be nice if we could call `vault.execute_procedure(...)` with procedures that only contain `VaultLocation`s and the `Location` is figured out from the vault context. I haven't looked into how procedures work in detail, but one way to do it may be by injecting the vault path into the `State` and requiring the procedures to call `State::location(&self, vault_location: VaultLocation) -> Location` in order to convert it. Admittedly, slightly more annoying to code. If procedures are contained to one vault, then "cross-vault" procedures would no longer be possible, at least not trivially. So if it was ever a feature, for example, to be able to read from vault1 and write into vault2, then this would not be a viable API.

Alternatively, the `execute_procedure` method can move to the `Client`, and work exactly like it currently does. API-wise it's a bit confusing to have `VaultLocation` *and* `Location` in the public interface, but it would work.

## Remote

This is the part I looked into the least, for time reasons. I believe most of the p2p management methods could probably work the same as it does now. There doesn't seem to be an explicit way to spawn an actor on a remote stronghold, so I'm unsure if a `RemoteClient`, `RemoteVault` and `RemoteStore` are feasible. In my opinion, it's fine if the remote and non-remote APIs do not exactly mirror each other. I would imagine, though, that a `RemoteClient` would be a nice encapsulation for policies. For example, a `Stronghold::remote_client` method would only return a `RemoteClient` instance if the current peer has access according to the policies.

Slightly related: The current remote methods like `write_remote_vault` are part of the `Stronghold` type and seem to be only additional features, but not standalone ones. Maybe I've missed something, but that wouldn't compile to Wasm, because it would also compile the entire stronghold library. Ideally, the library itself would also be feature-gated, and the p2p feature could be enabled by itself, so we could use a `RemoteStronghold` type that only requires `stronghold-p2p`.

### More Open Questions

- When I said the new interface has practically the same functionality as the current API, this is what I mean: There is this `former_client_path` feature in `read_snapshot`. I'm very unsure what this is useful for, but this is no longer present in the new API. `read_snapshot(client_path0, Some(client_path1), ...)` seems to simply act as if `read_snapshot(client_path1, None, ...)` had been called.

Most of what I propose can be built on top of the current interface, but with increased complexity and less performance. If nothing else, what I would _really_ like to see merged is the restore state functionality without requiring the password. Also, this is obviously not a ready-to-merge PR, as the P2P stuff is completely unaddressed at the moment. It is primarily meant to inspire discussion.

Disclaimer: I want to improve stronghold and figured instead of just giving feedback, I'll try to improve it myself, because I know you are busy with other things right now. I worked on this partly for fun, partly to learn more about stronghold internals, and partly to make life over in identity.rs and wallet.rs simpler. Merging this would be a significant change somewhat shortly before 1.0, so I can understand if that is considered too risky. Put another way: I don't expect this to be merged, but would be happy nonetheless, if it was.

Looking forward to any thoughts or comments!

## Type of change

- [ ] Documentation Fix

Added and modified tests in `interface_tests`.

## Change checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/client/src/actors.rs
+++ b/client/src/actors.rs
@@ -15,8 +15,10 @@ pub use self::{
 };
 pub use self::{
     registry::{
-        messages::{GetAllClients, GetClient, GetSnapshot, GetTarget, RemoveClient, SpawnClient, SwitchTarget},
-        Registry,
+        messages::{
+            GetAllClients, GetClient, GetSnapshot, GetTarget, RemoveClient, SetAllClients, SpawnClient, SwitchTarget,
+        },
+        Registry, Registry2,
     },
     secure::{messages as secure_messages, RecordError, VaultError},
     snapshot::{messages as snapshot_messages, returntypes as snapshot_returntypes},
--- a/client/src/actors/registry.rs
+++ b/client/src/actors/registry.rs
@@ -10,7 +10,7 @@
 
 use actix::{Actor, Addr, Context, Handler, Message, Supervised};
 use engine::vault::ClientId;
-use std::collections::HashMap;
+use std::{collections::HashMap, iter::FromIterator};
 
 #[cfg(feature = "p2p")]
 use super::p2p::NetworkActor;
@@ -67,6 +67,12 @@ pub mod messages {
 
     impl Message for GetAllClients {
         type Result = Vec<(ClientId, Addr<SecureClient>)>;
+    }
+
+    pub struct SetAllClients(pub Vec<(ClientId, Addr<SecureClient>)>);
+
+    impl Message for SetAllClients {
+        type Result = ();
     }
 }
 
@@ -207,5 +213,69 @@ impl Handler<p2p_messages::RemoveNetwork> for Registry {
         // Dropping the only address of the network actor will stop the actor.
         // Upon stopping the actor, its `StrongholdP2p` instance will be dropped, which results in a graceful shutdown.
         self.network.take()
+    }
+}
+
+#[derive(Default)]
+pub struct Registry2 {
+    clients: HashMap<ClientId, Addr<SecureClient>>,
+    snapshot: Option<Addr<Snapshot>>,
+    #[cfg(feature = "p2p")]
+    network: Option<Addr<NetworkActor>>,
+}
+
+impl Supervised for Registry2 {}
+
+impl Actor for Registry2 {
+    type Context = Context<Self>;
+}
+
+impl Handler<messages::SpawnClient> for Registry2 {
+    type Result = Addr<SecureClient>;
+
+    fn handle(&mut self, msg: messages::SpawnClient, _ctx: &mut Self::Context) -> Self::Result {
+        if let Some(addr) = self.clients.get(&msg.id) {
+            return addr.clone();
+        }
+
+        let addr = SecureClient::new(msg.id).start();
+        self.clients.insert(msg.id, addr.clone());
+
+        addr
+    }
+}
+
+impl Handler<messages::GetClient> for Registry2 {
+    type Result = Option<Addr<SecureClient>>;
+
+    fn handle(&mut self, msg: messages::GetClient, _ctx: &mut Self::Context) -> Self::Result {
+        self.clients.get(&msg.id).cloned()
+    }
+}
+
+impl Handler<messages::GetSnapshot> for Registry2 {
+    type Result = Addr<Snapshot>;
+
+    fn handle(&mut self, _: messages::GetSnapshot, _: &mut Self::Context) -> Self::Result {
+        self.snapshot.get_or_insert(Snapshot::default().start()).clone()
+    }
+}
+
+impl Handler<messages::GetAllClients> for Registry2 {
+    type Result = Vec<(ClientId, Addr<SecureClient>)>;
+
+    fn handle(&mut self, _: messages::GetAllClients, _: &mut Self::Context) -> Self::Result {
+        let empty = HashMap::default();
+        let clients = std::mem::replace(&mut self.clients, empty);
+
+        clients.into_iter().collect()
+    }
+}
+
+impl Handler<messages::SetAllClients> for Registry2 {
+    type Result = ();
+
+    fn handle(&mut self, msg: messages::SetAllClients, _: &mut Self::Context) -> Self::Result {
+        let _ = std::mem::replace(&mut self.clients, FromIterator::from_iter(msg.0));
     }
 }
--- a/client/src/actors/snapshot.rs
+++ b/client/src/actors/snapshot.rs
@@ -3,9 +3,7 @@
 
 #![allow(clippy::type_complexity)]
 
-use actix::{Actor, Handler, Message, Supervised};
-
-use std::path::PathBuf;
+use actix::{dev::MessageResponse, Actor, Handler, Message, Supervised};
 
 use engine::{
     snapshot,
@@ -31,6 +29,7 @@ pub mod returntypes {
     use super::*;
 
     /// Return type for loaded snapshot file
+    #[derive(MessageResponse)]
     pub struct ReturnReadSnapshot {
         pub id: ClientId,
 
@@ -44,12 +43,13 @@ pub mod returntypes {
 
 pub mod messages {
 
+    use crate::state::snapshot::SnapshotFile;
+
     use super::*;
 
     pub struct WriteSnapshot {
         pub key: snapshot::Key,
-        pub filename: Option<String>,
-        pub path: Option<PathBuf>,
+        pub snapshot_file: SnapshotFile,
     }
 
     impl Message for WriteSnapshot {
@@ -68,14 +68,32 @@ pub mod messages {
     #[derive(Default)]
     pub struct ReadFromSnapshot {
         pub key: snapshot::Key,
-        pub filename: Option<String>,
-        pub path: Option<PathBuf>,
+        pub snapshot_file: SnapshotFile,
         pub id: ClientId,
         pub fid: Option<ClientId>,
     }
 
     impl Message for ReadFromSnapshot {
         type Result = Result<returntypes::ReturnReadSnapshot, ReadError>;
+    }
+
+    #[derive(Default)]
+    pub struct ActorStateFromSnapshot {
+        pub id: ClientId,
+    }
+
+    impl Message for ActorStateFromSnapshot {
+        type Result = returntypes::ReturnReadSnapshot;
+    }
+
+    #[derive(Default)]
+    pub struct LoadFromDisk {
+        pub key: snapshot::Key,
+        pub snapshot_file: SnapshotFile,
+    }
+
+    impl Message for LoadFromDisk {
+        type Result = Result<(), ReadError>;
     }
 }
 
@@ -111,7 +129,7 @@ impl Handler<messages::ReadFromSnapshot> for Snapshot {
                 data: Box::new(data),
             })
         } else {
-            let mut snapshot = Snapshot::read_from_snapshot(msg.filename.as_deref(), msg.path.as_deref(), msg.key)?;
+            let mut snapshot = Snapshot::read_from_snapshot(msg.snapshot_file, msg.key)?;
             let data = snapshot.get_state(id);
             *self = snapshot;
 
@@ -123,11 +141,35 @@ impl Handler<messages::ReadFromSnapshot> for Snapshot {
     }
 }
 
+impl Handler<messages::LoadFromDisk> for Snapshot {
+    type Result = Result<(), ReadError>;
+
+    fn handle(&mut self, msg: messages::LoadFromDisk, _ctx: &mut Self::Context) -> Self::Result {
+        let snapshot = Snapshot::read_from_snapshot(msg.snapshot_file, msg.key)?;
+        *self = snapshot;
+
+        Ok(())
+    }
+}
+
+impl Handler<messages::ActorStateFromSnapshot> for Snapshot {
+    type Result = returntypes::ReturnReadSnapshot;
+
+    fn handle(&mut self, msg: messages::ActorStateFromSnapshot, _ctx: &mut Self::Context) -> Self::Result {
+        let data = self.get_state(msg.id);
+
+        ReturnReadSnapshot {
+            id: msg.id,
+            data: Box::new(data),
+        }
+    }
+}
+
 impl Handler<messages::WriteSnapshot> for Snapshot {
     type Result = Result<(), WriteError>;
 
     fn handle(&mut self, msg: messages::WriteSnapshot, _ctx: &mut Self::Context) -> Self::Result {
-        self.write_to_snapshot(msg.filename.as_deref(), msg.path.as_deref(), msg.key)?;
+        self.write_to_snapshot(msg.snapshot_file, msg.key)?;
 
         self.state = SnapshotState::default();
 
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -22,7 +22,7 @@ use crate::{
     procedures::{CollectedOutput, Procedure, ProcedureError},
     state::{
         secure::SecureClient,
-        snapshot::{ReadError, Snapshot, SnapshotFile, WriteError},
+        snapshot::{ReadError, Snapshot as SnapshotActor, SnapshotFile, WriteError},
     },
     utils::{LoadFromPath, StrongholdFlags, VaultFlags},
     Location,
@@ -398,21 +398,36 @@ impl Stronghold {
     }
 }
 
-/// Represents the in-memory version of an encrypted snapshot file. The [`Vault`] can be used to
-/// securely store key material and operate on said material. An unencrypted HashMap-like [`Store`]
-/// can be used to store miscellaneous data.
 // Note: This is no longer `Clone`, because `write_snapshot`
 // needs to operate on the stronghold in a mutually exclusive manner.
-pub struct Stronghold2 {
+pub struct Snapshot {
+    file: SnapshotFile,
     registry: Addr<Registry2>,
 }
 
-impl Stronghold2 {
-    /// Initializes a new Stronghold.
-    pub fn new() -> Self {
+impl Snapshot {
+    pub fn new(file: impl Into<SnapshotFile>) -> Self {
         let registry = Registry2::default().start();
 
-        Self { registry }
+        Self {
+            file: file.into(),
+            registry,
+        }
+    }
+
+    /// Convenience wrapper to create a named snapshot in a OS-dependent default directory.
+    pub fn new_named(name: impl AsRef<str>) -> Self {
+        Self::new(SnapshotFile::named(name.as_ref()))
+    }
+
+    /// Convenience wrapper to create a snapshot from a path.
+    pub fn new_path(path: impl Into<PathBuf>) -> Self {
+        Self::new(SnapshotFile::path(path.into()))
+    }
+
+    /// Convenience wrapper to create a named snapshot with a default name.
+    pub fn new_main() -> Self {
+        Self::new(SnapshotFile::default())
     }
 
     /// Returns the [`Client`] identified by `client_path`.
@@ -435,14 +450,10 @@ impl Stronghold2 {
         Ok(Client::new(client_id, secure_client, snapshot_actor))
     }
 
-    /// Loads the snapshot from disk identified by TODO. The given `key_data` is used to
+    /// Loads the snapshot from the file system. The given `key_data` is used to
     /// attempt decryption of the file. After this method is called, `key_data` should be zeroized.
     // TODO: Can we take owned key_data and zeroize ourselves?
-    pub async fn read_snapshot<T: Zeroize + AsRef<Vec<u8>>>(
-        &mut self,
-        key_data: &T,
-        snapshot_file: SnapshotFile,
-    ) -> StrongholdResult<Result<(), ReadError>> {
+    pub async fn read<T: Zeroize + AsRef<Vec<u8>>>(&mut self, key_data: &T) -> StrongholdResult<Result<(), ReadError>> {
         let mut key: [u8; 32] = [0u8; 32];
         let keydata = key_data.as_ref();
 
@@ -450,7 +461,12 @@ impl Stronghold2 {
 
         let snapshot_actor = self.registry.send(GetSnapshot {}).await?;
 
-        let result = snapshot_actor.send(LoadFromDisk { key, snapshot_file }).await?;
+        let result = snapshot_actor
+            .send(LoadFromDisk {
+                key,
+                snapshot_file: self.file.clone(),
+            })
+            .await?;
 
         if let Err(e) = result {
             return Ok(Err(e));
@@ -459,10 +475,10 @@ impl Stronghold2 {
         Ok(Ok(()))
     }
 
-    pub async fn write_snapshot<T: Zeroize + AsRef<Vec<u8>>>(
+    /// Write the snapshot to the path contained in the snapshot.
+    pub async fn write<T: Zeroize + AsRef<Vec<u8>>>(
         &mut self,
         keydata: &T,
-        snapshot_file: SnapshotFile,
     ) -> StrongholdResult<Result<(), WriteError>> {
         let mut key: [u8; 32] = [0u8; 32];
         let keydata = keydata.as_ref();
@@ -483,7 +499,12 @@ impl Stronghold2 {
         let previous_len = weak_addrs.len();
 
         // Don't short-circuit, as that would get rid of all clients.
-        let res = snapshot.send(WriteSnapshot { key, snapshot_file }).await;
+        let res = snapshot
+            .send(WriteSnapshot {
+                key,
+                snapshot_file: self.file.clone(),
+            })
+            .await;
 
         // It's important that the actor task/thread gets a chance to run, in order to
         // drop the actors whose references have now gone away.
@@ -514,11 +535,11 @@ impl Stronghold2 {
 pub struct Client {
     id: ClientId,
     secure_client: Addr<SecureClient>,
-    snapshot_actor: Addr<Snapshot>,
+    snapshot_actor: Addr<SnapshotActor>,
 }
 
 impl Client {
-    fn new(id: ClientId, secure_client: Addr<SecureClient>, snapshot_actor: Addr<Snapshot>) -> Client {
+    fn new(id: ClientId, secure_client: Addr<SecureClient>, snapshot_actor: Addr<SnapshotActor>) -> Client {
         Client {
             id,
             secure_client,
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -15,14 +15,14 @@ use crate::{
             CheckRecord, CheckVault, ClearCache, DeleteFromStore, GarbageCollect, GetData, ListIds, ReadFromStore,
             ReloadData, RevokeData, WriteToStore, WriteToVault,
         },
-        snapshot_messages::{FillSnapshot, ReadFromSnapshot, WriteSnapshot},
-        GetAllClients, GetClient, GetSnapshot, GetTarget, RecordError, Registry, RemoveClient, SpawnClient,
-        SwitchTarget,
+        snapshot_messages::{ActorStateFromSnapshot, FillSnapshot, LoadFromDisk, WriteSnapshot},
+        GetAllClients, GetClient, GetSnapshot, GetTarget, RecordError, Registry, Registry2, RemoveClient,
+        SetAllClients, SpawnClient, SwitchTarget,
     },
     procedures::{CollectedOutput, Procedure, ProcedureError},
     state::{
         secure::SecureClient,
-        snapshot::{ReadError, WriteError},
+        snapshot::{ReadError, Snapshot, SnapshotFile, WriteError},
     },
     utils::{LoadFromPath, StrongholdFlags, VaultFlags},
     Location,
@@ -31,7 +31,7 @@ use engine::vault::{ClientId, RecordHint, RecordId};
 #[cfg(feature = "p2p")]
 use p2p::{identity::Keypair, DialErr, InitKeypair, ListenErr, ListenRelayErr, OutboundFailure, RelayNotSupported};
 
-use actix::prelude::*;
+use actix::{prelude::*, WeakAddr};
 use serde::{Deserialize, Serialize};
 use std::{path::PathBuf, time::Duration};
 use thiserror::Error as DeriveError;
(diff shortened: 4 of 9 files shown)


tensor-programming:
I have a concerns regarding some of these changes.

### Interface

Number one looks like a decent change.  Honestly, I had thought awhile ago that it would have made more sense to stop at clients as the top level collection type rather then having `Strongholds`.  Having a single stronghold only really serves a purpose because of the way that Riker worked in the past, and letting users deploy as many clients as they need on their own would be more efficient.  

For number two, not sure what you mean here.  The method names are `read_from_store`, `write_to_store`, `write_to_vault` etc; their intent is very clear.  The `delete_data` method (the only vague one I can see) used to be called something like `revoke_from_vault`; not sure why that was changed to be honest. 

For number three, not sure why it was thought that storing a password in plain text was a good idea. Storing a password as plain text in memory is a very large security issue.  The way that the system was created, it was always intended to ask for the password only when the snapshot is being read from/written to.  A more effective way to deal with this is to use some kind of credential cache or to just clear out the password after using it and ask for it again when the snapshot state is empty. This is less of a stronghold problem and more of a problem with how the wallet and identity decided to implement the software.  Also, making it so that the stronghold only needs the password on startup is not a solution.  Stronghold could be used in long running processes, in which case, having the snapshot state in memory persistently opens up the stronghold to potential abuse should someone have access to the computer running the system after authentication. A better security idiom is to drop the snapshot state after its used and ask for the password every time the snapshot is read into memory.  It also doesn't make sense to keep the snapshot in memory when its already in memory in the state of the main stronghold system. The main purpose for the current implementation was to allow the system to sort the multiple clients from a snapshot into an easy to deal with setup.  

### Actor State:

Cloning actors like that is problematic.  As you've mentioned it, just opens up the system to potential race conditions. Rather then cloning the actor, an actor should always be queried from the registry.  As a side note, the original implementation of Stronghold used an Actor model for isolation purposes.  Actix doesn't really provide these methods of isolation and therefore only seems to create issues due to its inherent complexity.  If I was still leading the project, I would suggest removing the actor model entirely or looking for a library more similar to what riker provided as a replacement to actix.  It might seem silly to keep switching these libraries, but there doesn't seem to be an inherent upside for using the actix system anymore. In fact, when I wrote elixir bindings for Stronghold, actix had to be stripped out entirely because it caused conflicts with the BEAM. 

### Procedures:

Stronghold is composed of multiple "primitives".  There is the high level Stronghold primitive which is a collection of clients, the client primitive which is a collection of vaults, the vault primitive which is a collection of records which all use the same encryption key and then the record primitive which is where the data resides (among other things).  Each vault id corresponds to an encryption key in the keystore structure and the `client_id`, `vault_id` and `record_id` are basically keys used to find values in the vault. 

Abstracting away the vault for procedures would be problematic for a few reasons.  As you pointed out, it would no longer be trivially possible to preform procedures between multiple vaults. The vault path itself is used to get the encryption key from the keystore; removing the explicit vault path from a location type would remove the ability to do this for multiple vaults in that context.  Again, there is no real assumption on where the data should be placed in the stronghold, but making this change would force cryptographic data to be stored in a single vault which would make the security slight more brittle.  There also doesn't seem to be much of an upside for doing this from what I can see. Not only would the client need to be refactored, the procedure macros and the engine would need heavy refactoring to make this possible.  Procedures are meant to be plug and play, and this change removes some of that spirit. 


### Open Questions:

The former Client path was made to allow a user to move the internals of a client to another client. That's what this former client variable is for.  If its been changed since I last looked at the code, then I am not sure why though I am sure @elenaf9 or @felsweg-iota can shed some light on the reason. 

I like the idea of removing the stronghold type in principle since it only really serves the purpose of making the wallet easier to manage (each wallet account corresponds with a new client).  To some extent this hinders stronghold as a generic security library by putting some of its architectural bias towards that of the firefly wallet's structure.  Outside of that, most of these changes just don't seem appropriate for the system.

PhilippGackstatter:
Thanks for taking the time to respond in detail, much appreciated! For now, I have just a follow-up for clarification.

> A better security idiom is to drop the snapshot state after its used and ask for the password every time the snapshot is read into memory.

I did not actually change how long the state is held in memory. In current `dev`, the snapshot state is not dropped from the snapshot actor, as far as I can tell. I merely separated the `ReadFromSnapshot` actor messages in two parts, one for each branch of this if statement:

https://github.com/iotaledger/stronghold.rs/blob/129225efd7a1a17bfa20f4767b5086f6796b3218/client/src/actors/snapshot.rs#L106-L121

So, when reading an actor's state, the password is not needed when it is already in the snapshot actor's current state. If that is intended functionality, I think this should be exposed properly through two different functions. 

I thought this would need a change in the interface, but I realized this is already possible by first calling `read_snapshot` with a snapshot name and the proper password required for decryption. Subsequently, one can call `read_snapshot` with a random password, because the state is already loaded. While it works, it feels like a hack.

<details>
  <summary>Here is a test showing that.</summary>

```rust
#[actix::test]
async fn stronghold_read_state() {
    let password = b"abcdefghijklmnopqrstuvwxyz012345".to_vec();
    let path: PathBuf = "./strong.hodl".into();

    let client_path0 = b"test a";
    let client_path1 = b"test b";

    let vault_path0: &[u8] = b"123";
    let vault_path1: &[u8] = b"456";

    let mut stronghold = Stronghold::init_stronghold_system(client_path0.to_vec(), vec![])
        .await
        .unwrap();

    stronghold
        .write_to_vault(
            Location::generic(vault_path0, b"test".to_vec()),
            b"test".to_vec(),
            RecordHint::new(b"").unwrap(),
            vec![],
        )
        .await
        .unwrap()
        .unwrap();

    stronghold
        .spawn_stronghold_actor(client_path1.to_vec(), vec![])
        .await
        .unwrap();

    stronghold
        .write_to_vault(
            Location::generic(vault_path1, b"test2".to_vec()),
            b"secret1".to_vec(),
            RecordHint::new(b"").unwrap(),
            vec![],
        )
        .await
        .unwrap()
        .unwrap();

    stronghold
        .write_all_to_snapshot(&password, None, Some(path.clone()))
        .await
        .unwrap()
        .unwrap();

    std::mem::drop(stronghold);

    let mut stronghold = Stronghold::init_stronghold_system(client_path0.to_vec(), vec![])
        .await
        .unwrap();

    stronghold
        .spawn_stronghold_actor(client_path1.to_vec(), vec![])
        .await
        .unwrap();

    stronghold.switch_actor_target(client_path0.to_vec()).await.unwrap();

    stronghold
        .read_snapshot(client_path0.to_vec(), None, &password, None, Some(path.clone()))
        .await
        .unwrap()
        .unwrap();

    assert!(stronghold.vault_exists(vault_path0.clone()).await.unwrap());

    stronghold.switch_actor_target(client_path1.to_vec()).await.unwrap();

    assert!(!stronghold.vault_exists(vault_path1.clone()).await.unwrap());

    // Load client1's state, without requiring the actual password.
    stronghold
        .read_snapshot(
            client_path1.to_vec(),
            None,
            // Not the original password.
            &[0xff; 32].to_vec(),
            None,
            Some(path.clone()),
        )
        .await
        .unwrap()
        .unwrap();

    assert_eq!(
        stronghold
            .read_secret(vec![], Location::generic(vault_path1, b"test2".to_vec()))
            .await
            .unwrap()
            .unwrap(),
        b"secret1".to_vec(),
    );

    std::fs::remove_file(path).unwrap();
}
```
</details>

So if this is intended, I think this would better be exposed through two functions:

```rust
impl Stronghold {
  // This is the current method, unchanged.
  pub async fn read_snapshot<T: Zeroize + AsRef<Vec<u8>>>(
      &mut self,
      client_path: Vec<u8>,
      former_client_path: Option<Vec<u8>>,
      keydata: &T,
      filename: Option<String>,
      path: Option<PathBuf>,
  ) -> StrongholdResult<Result<(), ReadError>> { ... }
  
  // This attempts to read from the loaded snapshot, without requiring `key_data`,
  // returning false if `client_path` isn't loaded.
  pub async fn read_from_existing_snapshot<T: Zeroize + AsRef<Vec<u8>>>(
      &mut self,
      client_path: Vec<u8>,
      former_client_path: Option<Vec<u8>>,
  ) -> StrongholdResult<bool> { ... }
}
```

tensor-programming:
> I did not actually change how long the state is held in memory. In current dev, the snapshot state is not dropped from the snapshot actor.

I am aware of that.  What I was saying is that it is kind of an oversight to begin with to do this.  It duplicates the state needlessly and opens up a security flaw in the system.

tensor-programming:
I thought about this a bit more over the past couple of days.  

I suspect one of the reasons why I made a choice to keep the state in memory was to make it lazy with regards to multiple clients. For example, in firefly, each client corresponds with an account which means a user could open one account without opening the others.  But this really only made sense in the context of the original actor model where in each client actor also was associated with 2 other actors (each account had 3 actors).  Back in that version, it made sense to not spawn all of the actors at one time and instead just spawn them as needed.  But with actix, this doesn't make much sense anymore.  Yes, each client is its own separate actor still, but a single client actor is not expensive to keep open as it was with 3 riker actors.  

My belief is that the snapshot state should be cleared after reading and after writing. In other words, it should only be populated when you are reading from or writing to a snapshot.  This means that the user should be prompted for the password on read and on write.  @elenaf9's write to snapshot with a password in a vault message makes it possible to remove that restriction for writes which is fine as an optional feature.  However, it should never be possible to read a snapshot without providing a password because of the security risks associated with such a move.  

As it is right now, there is no real benefit to the changes proposed in this regard because you can keep the snapshot open (aka keep the state in the app) as long as you want while still writing down to the snapshot without having to rely on the snapshot state to refresh the state.  Essentially what you are proposing can be done without the changes and it maintains a certain level of security without having to duplicate the state for no real purpose.

PhilippGackstatter:
> My belief is that the snapshot state should be cleared after reading and after writing. In other words, it should only be populated when you are reading from or writing to a snapshot.

Let's say you read a snapshot that contains data from 3 clients. You only load the state into 2 clients. Now the state of 2 clients is in memory, and the state of the third client has been dropped from memory. Now when you write back to the snapshot, you would overwrite the snapshot and effectively erase the third client's state, iiuc. Isn't that the actual reason why the snapshot state is being kept in memory? So there are three options, of which 2 and 3 solve that overwriting problem:

1. Keep the snapshot state in memory, and load and store as needed. (Current implementation).
2. When loading a snapshot, spawn an actor for each client and load that client's state from the snapshot state, then drop the snapshot state. This seems wasteful for snapshots containing many clients.
3. Best explained with an example: Load the snapshot, e.g. for three clients. The user wants to work with client 1 and 2, so copy the state from the snapshot into those two actors. On write, read the snapshot again, which populates the snapshot state with *all* client's state. Then retrieve client 1 and 2's state and overwrite those entries in the snapshot state. Client 3 in the snapshot state is unchanged. Write to disk.

Option 3 means you need the password on write, in order to read, which, if I understand you correctly, should not be possible without the user providing that password. So with 294, that PR would either use the password from the vault to read the snapshot, or the PR becomes obsolete, because the password is required for writing (in order to read).

I think you are implying option 2. But I think if maximum security is the goal, no matter the cost to users, option 3 would be best, to minimize the time that keys are in memory?

> As it is right now [...] you can keep the snapshot open (aka keep the state in the app) as long as you want

It seems to me, the assumption is that holding the snapshot state in memory is bad, but having running clients is considered more or less okay. Why? Don't they have the same security guarantees, since both states are memory-protected in the same way? In which case, the only bad thing is that you have the same keys protected in memory twice, instead of once, which isn't ideal, but doesn't seem like that big of a problem. Especially when that option makes the library user's life harder. It means that the password needs to be passed into stronghold much more often, and each time, it potentially passes through multiple layers of the application in unprotected memory, likely even non-Rust languages, where it's much easier for someone to forget to zeroize the password. Honestly, though, not sure how valid that argument is. So the question is: Would a credential cache/keyring on application level to cache/store passwords be an actual better design, considering the security across the entire application stack?


Finally, I think even with the current implementation it's possible to have a high level of security, even more with a refactored interface that has a dedicated `Snapshot` type. (Not yet part of the proposal). A maximum security application might want to only hold snapshot, client and vault in memory for the shortest amount of time possible, e.g. only while doing a signing operation. For every operation, they would need the password. However, applications that want to strike more of a balance between usability and security might be fine with holding the snapshot, client and vault in memory for 5 minutes, before dropping the snapshot and again requiring the password for the next operation. But during those 5 minutes, as a user of stronghold, it's much easier to work with, because no password handling or prompting is required. Applications can make the decision when to drop the snapshot in their architecture (which is harder to get wrong than zeroizing passwords in, say, JavaScript). Do you think that's a valid approach?

felsweg-iota:
Thanks @PhilippGackstatter for providing a decent proposal and ideas for Stronghold's system improvement. Also thanks to @tensor-programming providing quick feedback on the topics mentioned. I would like to comment on the proposed features and overhaul of the major type system inside the Stronghold type system, as I also see a huge benefit to make the use of Stronghold much easier, while providing a decent security. In this regard, I would also like to add a few more points where we see how the current architecture can be improved in terms of convenience and security after all. I also want to make sure, that the current implementation is a migration from the old riker based system to actix. The functionality is the same plus the extension we have build on top of it.  


## Feedback 

I will only add points, that I think might be worth discussing, if they haven't been already addressed earlier. 

### Interface

 I agree, this seems to be a good proposal to work more on the type level, and make the underlying actor system working transparently. The question remains, how we would solve concurrency problems when working on different client path at almost the same time. 
   Sticking one client to a vault and store would render the currently flexibility to chose a target obsolete, and would make it more or less rigid, unless there is a reason to reduce the number of parameters to provide a more lean call of the underlying functionality. 
 
Feedback on the Snapshot can be found below. 


### Actor System

Having [WeakAddr](https://docs.rs/actix/0.12.0/actix/struct.WeakAddr.html)esses and strong [Addr](https://docs.rs/actix/0.12.0/actix/struct.Addr.html)esses (eg. dropping it, would terminate the referenced actor) doesn't necessarily solve the problem of race conditions in this sense, as each `Client` would request operations (eg. send messages) from the underlying actor, that are internally multiplexed. It also seems to circumvent the problem the current implementation has with bundling atomic transactions. 
The second proposal sounds a bit like some type of garbage collection for clients, that again tries to circumvent the shortcomings of the actix actor system. 

I agree with @tensor-programming in this regard, that actix doesn't provide any more value, that just providing a more abstracted concurrent system, that runs on one system-thread, multiplexing all futures on a very powerful event loop. 

A system with memory isolation, preventing certain system calls (like printing etc) would be desirable. Another feature to have inside a Stronghold would be to provide some kind of multi-tenancy in an isolated manner. Such an architecture would also abstract away local and remote calls, each with their security promises given; let the calling clients conduct bundled atomic transactions, that can be commited or rolled-back anytime. 

### Cryptographic Procedures

As one of the major features of the `proc-api` is it's inherent composability of complex cryptographic functions out of primitive ones. I agree with @tensor-programming here, in that introducing a strong coupling between vaults and procedure the original intent would be gone. One major advantages is to work on multiple vaults out of composabilty. 

### Remote

As you already pointed it out, it is not possible to spawn a remote client. An upcoming [PR](https://github.com/iotaledger/stronghold.rs/pull/292) introduces a policy engine to manage access with fine grained controls. Ideally the policies are hierarchical by nature. This PR also introduces the possibility to map peers to clients. One can think of some standard policy with access to certain secrets that can be applied to a default client, that can be spawned by the network actor on default. This arises a few questions:

- How is the configuration done on remote side? Does someone needs to login to configure the stronghold system integrated into some application? 
- How would the default actor react to dynamic changes? A requesting peer could be allowed to write certain secrets, a policy could be set to enable a more coarse access to secrets (eg. running cryptographic procedures, derive public keys etc..)

### Open Questions

> The former Client path was made to allow a user to move the internals of a client to another client. That's what this former client variable is for. If its been changed since I last looked at the code, then I am not sure why though I am sure @elenaf9 or @felsweg-iota can shed some light on the reason.

Just to be clear on this point: This functionality is still present in the current `dev` branch. The `former_client_path` can be used to derive a previously used client to load the state. The former implementation passed the calls to read from a snapshot down to the snapshot actor, that loads state for the current client_id or the former_client_id see [here](https://github.com/iotaledger/stronghold.rs/blob/37d7d5371e925c6846f7e87f5a3705dc579ccfbc/client/src/actors/snapshot.rs#L94). The migrated and modified implementation with actix does the same, but on a higher level. The former_client_id is [still in use](https://github.com/iotaledger/stronghold.rs/blob/129225efd7a1a17bfa20f4767b5086f6796b3218/client/src/actors/snapshot.rs#L104). 
 
 
### Snapshot

- The original intend of ReadSnapshot with two Options of a path or file_name respectively, was to load a file from an explicit path, or load it from a default directory. This [change](https://github.com/iotaledger/stronghold.rs/blob/037df4a0eda83ad51d1cce79f0114a7e0df9edcd/client/src/actors/snapshot.rs#L148) would force a user to always provide a path. IMHO it would be better to split the original function into two: Having a separated function for each intent eg. `load_from_path` would load a snapshot, while `load_default` would stick to a configur(able|ed) default path to a snapshot file. 

## Proposals

The following two subsections are proposals to encounter the problems described above on a more fundamental level. I put them here open for debate, but do not reflect mandatory changes, that will necessarily be integrated in a next release of Stronghold.

### System-Architecture

As we are experiencing now, employing actix as a single (system) threaded actor system does not provide much benefit when it comse to actor/state isolation and security options. But it manages concurrency in itself quite well; the interfaces are clear and simple, though. One very negative point about actix is the "consumption" of an underlying tokio runtime, that renders Stronghold using in a shared tokio context out of scope; the problems described by @PhilippGackstatter above are a direct result of this architectural decision. 

I can envision a system architecture that is concurrent by nature, can be run in a shared (threaded) context, having the same security guarantees and even more provides some concepts taken from a secure actor, that supresses any security related system calls, and makes requests transparent on a local and on a remote level. Furhtermore, since concencurrency is at high stake, operations on the underlying vaults can be done in an atomic transactions: bundled operations, that will block any other operation during their execution. Such a system could be build with a [software transactional memory](https://en.wikipedia.org/wiki/Software_transactional_memory) based architecture.

### Snapshot 

The snapshot system can be reworked in such a way, that writes into the snapshot file could be seen as commits. Each write would have a transactional number, that can restore a specific state. This would be a versioned snapshot file. The integration could be done in such a way, that multiple readers need to be developed to be downwards compatible. Clients of the newer version would be able to load snapshot states at specific commits, or will be returned always the head commit of the snapshot file. This api must be optional by nature, as older versions must be supported. 

As already pointed out, the current implementation allows to load a specific snapshot state from memory, once a snapshot file has been read. One option to add an additional layer of security, and to reduce the tear and wear on persistent memory drives could be to re-rencrypt the snapshot state in memory, will storing a key inside the vault to re-open the encrypted snapshot in memory. This would not expose any passwords inside insecure memory areas.

tensor-programming:
> As we are experiencing now, employing actix as a single (system) threaded actor system does not provide much benefit when it comse to actor/state isolation and security options. But it manages concurrency in itself quite well; the interfaces are clear and simple, though. One very negative point about actix is the "consumption" of an underlying tokio runtime, that renders Stronghold using in a shared tokio context out of scope; the problems described by @PhilippGackstatter above are a direct result of this architectural decision.

Agreed, these are my thoughts as well atm.  

>I can envision a system architecture that is concurrent by nature, can be run in a shared (threaded) context, having the same security guarantees and even more provides some concepts taken from a secure actor, that supresses any security related system calls, and makes requests transparent on a local and on a remote level. Furhtermore, since concencurrency is at high stake, operations on the underlying vaults can be done in an atomic transactions: bundled operations, that will block any other operation during their execution. Such a system could be build with a software transactional memory based architecture.

Didn't consider STM but it would be kind of perfect here @felsweg-iota great suggestion.

>The snapshot system can be reworked in such a way, that writes into the snapshot file could be seen as commits. Each write would have a transactional number, that can restore a specific state. This would be a versioned snapshot file. The integration could be done in such a way, that multiple readers need to be developed to be downwards compatible. Clients of the newer version would be able to load snapshot states at specific commits, or will be returned always the head commit of the snapshot file. This api must be optional by nature, as older versions must be supported.

I would be very careful with this, but the idea of commits and basically having a version controlled structure in the snapshot would work well in this regard.  Like you said though, we always have to support the other format. 

> As already pointed out, the current implementation allows to load a specific snapshot state from memory, once a snapshot file has been read. One option to add an additional layer of security, and to reduce the tear and wear on persistent memory drives could be to re-encrypt the snapshot state in memory, will storing a key inside the vault to re-open the encrypted snapshot in memory. This would not expose any passwords inside insecure memory areas.

I do like this idea in theory but let me kind of throw in my two cents here.  

Thinking about the security issue, the real security holes in stronghold come from the procedures. Specifically, we are talking about procedures that add or remove some thing from the vault or return something derived from a secret in the vault.  Reading a snapshot does nothing if the user can't access any procedures and our memory guards are working as they should.  Even writing a snapshot doesn't matter either if there is no modification of the internal data.  And in many of the use cases I can think of; we are talking about strongholds that only have a couple of pieces of data/vaults.  In these cases, the main concern comes back to those "write" procedures.   Another thing to keep in mind is that most implementations will immediately want to write to the snapshot after something changes in the state.  I know this is how things work in firefly and in some of the other systems I have used.  

My thoughts are that we could keep a policy that follows the original format of requiring a password on snapshot read and on snapshot write (with the option of holding the password in a temporary record to make writing easier). And then create a policy around locking the runtime rather then locking the snapshot.

felsweg-iota:
Solved by [PR#324](https://github.com/iotaledger/stronghold.rs/pull/324)

PR #297, opened 2021-11-29
feat(client): add procedure for copying a record

### LINKED ISSUE #283: Enable moving keys
## Description

Add a function to move a key from one location to another. This way we can generate a key at some random location `xyz`, use the some part of the key (e.g. public key) to generate the actual location `location`, and then move the key from `xyz` to `location`.

## Motivation

We would like to have this in identity.rs for two reasons. 

1. We would like to use a DID of an identity in the locations for everything that is stored for that identity. The DID itself is the hash of the public key of the initial keypair. So if we want to store the initial keypair at a location that is based on the DID, we have a circular dependency. Currently, we generate the keypair in-memory to work around this limitation, but this is less secure than being able to generate it in Stronghold directly.

2. When creating a new verification method `key-1`, we generate a new key in stronghold. If we later want to update the key material of the method `key-1`, we cannot use a location that is only based on the unique `key-1` identifier. This is because we may need the old key in order to sign a new document update, and only then could we get rid of the old key. Our current solution is to keep some extra state which is part of the location. We might move to a new approach where the public key together with the fragment is used as the location, so that two `key-1`'s with different underlying key materials yield two different locations. That way, we don't have to keep the additional state, reducing complexity.

## Requirements

Add a method along the lines of 

```rust
Stronghold::move_data(&self, current_location: Location, new_location: Location);
```

that moves the key (names are suggestions, feel free to change).

No.

elenaf9:
# Description of change

Add procedure for copying a secret from one location to another.
See #283 for more info; cc @PhilippGackstatter.

Fixes issue #283 

## Type of change

Added test `procedure_tests::usecase_move_record`.

## Change checklist

### DIFF
--- a/.changes/client_procedures_copy-record.md
+++ b/.changes/client_procedures_copy-record.md
@@ -1,0 +1,5 @@
+---
+"iota-stronghold": patch
+---
+
+- [[PR 297](https://github.com/iotaledger/stronghold.rs/pull/297)] Add `CopyRecord` procedure.
--- a/client/src/procedures.rs
+++ b/client/src/procedures.rs
@@ -5,9 +5,9 @@ mod primitives;
 mod types;
 
 pub use primitives::{
-    AeadAlg, AeadDecrypt, AeadEncrypt, BIP39Generate, BIP39Recover, Chain, ChainCode, Ed25519Sign, GenerateKey, Hash,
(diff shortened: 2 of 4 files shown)


felsweg-iota (approved):
Looks good to me!

Issue #268, opened 2021-10-13
Panic When Trying to Insert Many Keys into Vaults


HenriqueNogara:
# Description
The code attached to this issue causes the following error: 
```
thread '<unnamed>' panicked at 'Error setting memory protection to NoAccess', /home/henrique/.cargo/git/checkouts/stronghold.rs-fe28308a6634692d/aea8a9d/engine/runtime/src/boxed.rs:289:9
```
Here is my OS info: 
```
OS: Ubuntu 20.04.3 LTS
OS Type: 64 bit
Gnome Version: 3.36.8
Windowing System: X11
```

The error is always after 8173 keys

code: https://gist.github.com/HenriqueNogara/f2caa968abf03ef860ead62d31a20cea

felsweg-iota:
system configuration:
```
OS: Ubuntu 20.04.3 LTS
OS Type: 64 bit
Gnome Version: 3.36.8
Windowing System: X11
```

This seems to be a system's restriction. [libsodium](https://doc.libsodium.org/memory_management) explicitly states, that some systems enforce a limit on possible memory protections:

> Note: Many systems place limits on the amount of memory that may be locked by a process. Care should be taken to raise those limits (e.g. Unix ulimits) where neccessary. sodium_mlock() will return -1 when any limit is reached.

Taking a deep dive into the [memory protection API](https://linux.die.net/man/2/mlock) concerning linux, non-priviledged processes are constraint in number of protected memory regions. I could successfully recreate the error on my system (same config).
Since libsodium matches the produced error stored in `errno`, I could verify that return `errno` is `ENOMEN`, that is described as

> the caller had a nonzero RLIMIT_MEMLOCK soft resource limit, but tried to lock more memory than the limit permitted.

<details>
<summary>Here is a minimal variation of the test case mentioned above</summary>

```lang:rust

use engine::vault::RecordHint;
use stronghold_utils::random as rand;

#[actix::test]
async fn test_maximum_vault_entries() {
    let stronghold = Stronghold::init_stronghold_system(b"client_path".to_vec(), vec![])
        .await
        .unwrap();

    for i in 0..usize::MAX {
        let location = Location::Generic {
            vault_path: rand::bytestring(32),
            record_path: rand::bytestring(32),
        };

        let hint = RecordHint::new(rand::bytestring(24)).unwrap();
        let payload = rand::bytestring(56);

        if let Err(error) = stronghold.write_to_vault(location, payload, hint, vec![]).await {
[...]

PR #301, opened 2021-12-15
Client: document restriction for the amount of used vaults

elenaf9:
# Description of change

Follow up on issue #268: for now add a comment to the code to document that there is an upper bound for the number of vaults that can be used. Eventually it should check during runtime if the upper bound of locked memory is reached **before** trying to lock additional memory.

Apart from that add some smaller fixes more `procedure_tests` and deals with clippy v0.1.57.

## Type of change

- [x] Documentation Fix


felsweg-iota (approved):
Looks good to me

PhilippGackstatter (approved):
Looks good. Thanks for the extended documentation!

PR #302, opened 2021-12-15
P2p: add examples and improve docs

elenaf9:
# Description of change

Add two examples:
1. Demonstrate how the stronghold-p2p firewall may be used to set dynamic rules for each individual peer once they connect, and restrict permitted requests based on the type of request.
2. Showcase how a standalone stronghold-p2p peer (without a local Stronghold) can interact with a remote Stronghold through p2p. cc @PhilippGackstatter

Enhance docs.

_Note: This PR cherry-picks 420ecb04b07b055109539fda2fb877712cc98597 and fe7867b0135a5b0678e0d1a39c710818507aa704 from #301 to fix the checks._

## Type of change

- [x] Documentation Fix

### DIFF
--- a/p2p/examples/dynamic-firewall.rs
+++ b/p2p/examples/dynamic-firewall.rs
@@ -97,7 +97,7 @@ fn allow_only_ping(request: &RequestPermission) -> bool {
 // - `-p <peer-id>`: Send a ping
 // - `-p <peer-id> -m <message>`: Send a message
 async fn on_user_input(
-    stronghold: &mut StrongholdP2p<Request, Response, RequestPermission>,
+    network: &mut StrongholdP2p<Request, Response, RequestPermission>,
     input: String,
 ) -> Result<(), Box<dyn Error>> {
     let peer_regex = "-p\\s+(?P<target>[[:alnum:]]{32,64})";
@@ -126,7 +126,7 @@ async fn on_user_input(
         None => Request::Ping,
     };
     // Send request and wait for response
-    match stronghold.send_request(peer, request).await {
+    match network.send_request(peer, request).await {
         Ok(res) => match res {
             Response::Pong => println!("Pong"),
             Response::Message(msg) => println!("Response: {}", msg),
@@ -256,15 +256,18 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Channel through which approved inbound requests are forwarded.
     let (request_tx, mut request_rx) = EventChannel::new(10, ChannelSinkConfig::Block);
 
-    let mut stronghold = StrongholdP2p::new(firewall_tx, request_tx, None).await?;
-    stronghold.start_listening("/ip4/0.0.0.0/tcp/0".parse()?).await?;
-    println!("Local Peer Id: {}", stronghold.peer_id());
+    let mut network = StrongholdP2p::new(firewall_tx, request_tx, None).await?;
+
+    network.start_listening("/ip4/0.0.0.0/tcp/0".parse()?).await?;
+    println!("\nLocal Peer Id: {}", network.peer_id());
+
+    println!("\nPing or message a remote peer:\n`-p <peer-id>`\t\t\t# Send a ping\n`-p <peer-id> -m <message>`\t# Send a message\n");
 
     let mut stdin = BufReader::new(stdin()).lines();
     loop {
         futures::select! {
             stdin_input = stdin.next_line().fuse() => match stdin_input? {
-                Some(line) => on_user_input(&mut stronghold, line).await?,
+                Some(line) => on_user_input(&mut network, line).await?,
                 None => break,
             },
             request = request_rx.select_next_some() => {
--- a/p2p/examples/dynamic-firewall.rs
+++ b/p2p/examples/dynamic-firewall.rs
@@ -103,9 +103,11 @@ async fn on_user_input(
     input: String,
 ) -> Result<(), Box<dyn Error>> {
     let peer_regex = "-p\\s+(?P<target>[[:alnum:]]{32,64})";
-    let peer: PeerId = match Regex::new(peer_regex).expect("Valid regex").captures(&input) {
+    let peer: PeerId = match Regex::new(peer_regex).expect("Valid regex.").captures(&input) {
         Some(capture) => {
-            let target = capture.name("target").expect("Target was captured.");
+            let target = capture
+                .name("target")
+                .expect("Regex should only match strings with a 'target' capture group.");
             match PeerId::from_str(target.as_str()) {
                 Ok(id) => id,
                 Err(_) => {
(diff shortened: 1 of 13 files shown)


p2p/examples/stronghold.rs
```diff
@@ -0,0 +1,142 @@
...
+    let (info_tx, info_rx) = oneshot::channel::<(PeerId, Multiaddr)>();
+    let local = async {
+        let (stronghold_id, stronghold_addr) = info_rx.await.unwrap();
+        let key_location = Location::generic("v0", "r0");
+        // Write a new key into the remote vault
+        local_client::generate_key(stronghold_id, stronghold_addr.clone(), key_location.clone())
+            .await
+            .unwrap();
+        // Use the generated key to sign multiple messages.
+        // In this example it is spawning a new client on each operations to demonstrate that they are stateless.
```
PhilippGackstatter: But the recommended approach is to setup a `StrongholdP2p` instance once and reuse it, correct?
If that is the recommended approach, I would prefer to have an example that showcases that, but I'm also happy with the current example and a note that states what the recommendation is.

p2p/examples/stronghold.rs
```diff
@@ -0,0 +1,142 @@
...
+#[actix_rt::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let (info_tx, info_rx) = oneshot::channel::<(PeerId, Multiaddr)>();
+    let local = async {
+        let (stronghold_id, stronghold_addr) = info_rx.await.unwrap();
+        let key_location = Location::generic("v0", "r0");
+        // Write a new key into the remote vault
+        local_client::generate_key(stronghold_id, stronghold_addr.clone(), key_location.clone())
+            .await
+            .unwrap();
```
PhilippGackstatter:
```suggestion
            .await?;
```

p2p/examples/stronghold.rs
```diff
@@ -0,0 +1,142 @@
...
+        // Use the generated key to sign multiple messages.
+        // In this example it is spawning a new client on each operations to demonstrate that they are stateless.
+        for i in 0..3 {
+            let message = format!("message {}", i);
+            local_client::sign_message(stronghold_id, stronghold_addr.clone(), key_location.clone(), message)
+                .await
+                .unwrap();
```
PhilippGackstatter:
```suggestion
                .await?;
```

PhilippGackstatter (approved):
Thanks for adding the p2p example!

Nit: Maybe you can rename the file to `remote_stronghold` or `p2p_stronghold` so it's a bit less ambiguous what `cargo run --example stronghold` does.

p2p/examples/stronghold.rs
elenaf9: I stayed with the current implementation but added a comment.

p2p/examples/dynamic-firewall.rs
```diff
@@ -0,0 +1,282 @@
...
+// - `-p <peer-id>`: Send a ping
+// - `-p <peer-id> -m <message>`: Send a message
+async fn on_user_input(
+    network: &mut StrongholdP2p<Request, Response, RequestPermission>,
+    input: String,
+) -> Result<(), Box<dyn Error>> {
+    let peer_regex = "-p\\s+(?P<target>[[:alnum:]]{32,64})";
+    let peer: PeerId = match Regex::new(peer_regex).expect("Valid regex").captures(&input) {
+        Some(capture) => {
+            let target = capture.name("target").expect("Target was captured.");
```
felsweg-iota:
```suggestion
            let target = capture.name("target").expect("Capture group not found.");
```
minor: it's rather "capture group not found"

p2p/examples/dynamic-firewall.rs
```diff
@@ -0,0 +1,282 @@
...
+        }
+        None => {
+            println!("Please include a target via \"-p <peer-id>\"");
+            return Ok(());
+        }
+    };
+    let type_regex = "-m (?P<msg>\"[^\"]*\"|\\S+)";
+    let request = match Regex::new(type_regex).expect("Valid regex").captures(&input) {
+        Some(capture) => {
+            let msg = capture.name("msg").expect("Message was captured.").as_str();
```
felsweg-iota:
```suggestion
            let msg = capture.name("msg").expect("Capture group not found").as_str();
```

p2p/examples/dynamic-firewall.rs
```diff
@@ -0,0 +1,282 @@
...
+                        let rule: Rule<RequestPermission> = Rule::Restricted {
+                            restriction: allow_only_ping,
+                            _maker: PhantomData,
+                        };
+                        rule_tx.send(FirewallRules::new(Some(rule), None)).unwrap();
+                        break;
+                    }
+                    _ => {
+                        tries += 1;
+                        if tries < 3 {
```
felsweg-iota: better put a constant outside

p2p/src/behaviour.rs
```diff
@@ -200,6 +200,50 @@ where
...
+    ///     None,
+    ///     Some(relay_behaviour),
+    ///     firewall_tx,
+    ///     FirewallConfiguration::default(),
+    ///     None
+    /// );
+    /// let swarm = Swarm::new(boxed_transport, behaviour, peer_id);
+    /// # Ok(())
+    /// }
+    /// ```
```
felsweg-iota, line 246: better remove the commented out code

felsweg-iota (changes requested):
Looks good to me!
Please take a look at the requested changes :)

p2p/examples/dynamic-firewall.rs
elenaf9: I personally like to use the message in `expect` to describe the expected invariant (= why to we expect it to be safe to unwrap here) rather than an error message that is just stating the obvious. Ideally a library/ application should never unwrap values unless they know that the value is `Ok`/ `Some`, so the message should never be shown to the user.

However, I agree that in this case the wording is not ideal. What do you think about something along the lines of 
`.expect("regex should only match strings with a 'target' capture group.").`
Alternatively we could also just use `unwrap` here and add a comment on why it is safe to unwrap.

p2p/examples/dynamic-firewall.rs
elenaf9: See comment above.

p2p/examples/dynamic-firewall.rs
elenaf9: Good idea! Done in 7b94aa3.

p2p/examples/dynamic-firewall.rs
felsweg-iota: > "regex should only match strings with a 'target' capture group."

I think that's a good choice.

elenaf9: Done in 05df78f.

felsweg-iota (approved):
Looks good to me!

PR #312, opened 2022-01-31
fix(client): fix clippy::to_string_in_format_args

elenaf9:
# Description of change

Fix cargo clippy lint `to_string_in_format_args` that was added in clippy v0.1.58.


felsweg-iota (approved):
LGTM

PR #315, opened 2022-02-08
feat/async software transactional memory

felsweg-iota:
# Description of change

This (draft) PR brings in a asynchronous software transactional memory implementation.

*notes*:
there are still some quirks, that need to be solved; nested transaction are not possible right now; memory operations are still plain, integration of `LockedMemory` is ongoing; 

to be defined.

## Type of change

Integration and module tests have been provided.

## Change checklist


stm/src/boxedalloc.rs
```diff
@@ -0,0 +1,116 @@
...
+impl<K> CompositeKey<K>
+where
+    K: AsRef<Vec<u8>> + Zeroize + Clone,
+{
+    /// Initializes the internal key fragments, consume the
+    /// provided actual key. The key must have a specific size,
+    /// depending on the underlying hashing algorithm.
+    pub fn new(key: K) -> Result<Self, KeyError> {
+        Ok(Self {
+            x: Arc::new(Mutex::new(key.clone())),
```
tensor-programming: Since this is a specific sized key type, I wonder can we replace the vector with a fixed size slice and avoid the clone here?  Maybe use the new rust const generic syntax to make something very easy to use.  I know its just an stand in at this point, just curious.

stm/src/transaction.rs
```diff
@@ -0,0 +1,299 @@
...
+        Arc, Mutex,
+    },
+    task::{Context, Poll},
+};
+
+/// Defines a transaction strategy. Strategies vary
+/// how transactions should be handled in case the commit
+/// to memory fails.
+#[derive(Clone, Copy)]
+pub enum Strategy {
```
tensor-programming: I like the idea of an exposed strategy API.  Makes sense in the context of the Stronghold runtime.  I wonder if its worth adding a strategy for dealing with runtime "DDOSing" (lack of a better term) where a user calls a procedure in perpetuity so they can attempt to dump the memory while the secrets are revealed.  Would it make sense to add that here or make it a procedure level API?

stm/src/types.rs
```diff
@@ -0,0 +1,224 @@
...
+            _ if a < b => Ordering::Less,
+            _ => Ordering::Equal,
+        }
+    }
+}
+
+impl<T> Eq for TVar<T> where T: Send + Sync + BoxedMemory {}
+
+/// Transactional Log type. The intend of this type
+/// is to track each operation on the target value
```
tensor-programming: this log would be encrypted right?  Don't remember what decision we came to on that one.

tensor-programming:
Very nice start, I like how you've set things up with this implementation of STM.  Look forward to seeing how it progresses forward in the future.

stm/src/boxedalloc.rs
felsweg-iota: True that, and this part will definitely not land in the final version. The `boxedalloc` module shall and will be superseded by the upcoming [`runtime` changes ](https://github.com/iotaledger/stronghold.rs/pull/316). This module is just a stub to allow implementation on interface level at least. Sorry, I should have made that clear :sweat_smile:

stm/src/transaction.rs
felsweg-iota: That's actually a good idea to tackle this at stm level, having some kind of execution throttling. This could be done by queuing up a limited number of transactions, blocking any further requests,

stm/src/types.rs
felsweg-iota: yes. currently working on a solution how to do this transparently.

engine/new_runtime/src/memories/file_memory.rs
```diff
@@ -0,0 +1,227 @@
...
+        fs::read(&self.fname)
+    }
+
+    // TODO: add security
+    // - file permissions
+    // - noise in the file
+    fn write_to_file(payload: &[u8]) -> Result<String, std::io::Error> {
+        let fname: String = FileMemory::<P>::random_fname();
+        let mut file = File::create(&fname)?;
+        file.write_all(payload.as_bytes())?;
```
nothingismagick, line 58: This is an amazing Pull Request. My only comment is that you might like to consider writing to a temporary file and then mv'ing it to its final name and performing a sync. Back in the day we saw some corruption with snapshots, and this is the way it was solved:

https://github.com/iotaledger/stronghold.rs/blob/1dfd61b9c393675173bb2e24a6e5b04d2a4bc473/engine/src/snapshot/logic.rs#L190-L196

nothingismagick (commented):
Just awesome stuff here.

engine/new_runtime/src/memories/file_memory.rs
felsweg-iota: Hey @nothingismagick, thanks for the nice words, but you are actually referring work from this [PR#316](https://github.com/iotaledger/stronghold.rs/pull/316). @vuongDang deserve the credit for implementing all the memory features :blush: 

I am only pulling in work from this branch to build on top of it.

felsweg-iota:
This PR will be closed and re-opened on another branch.

PR #316, opened 2022-02-08
Feat/locked memory

vuongDang:
[Link](https://github.com/iotaledger/stronghold.rs/blob/feat/locked-memory/engine/new_runtime/README.md) to the README.md


engine/new_runtime/src/types.rs
```diff
@@ -0,0 +1,48 @@
...
+        impls!{array $type; $size}
+    )*};
+
+    (prim $type:ty) => {
+        unsafe impl Bytes for $type {}
+    };
+
+    (array $type:ty; ($($size:tt)*)) => {$(
+        #[allow(trivial_casts)]
+        unsafe impl Bytes for [$type; $size] {}
```
felsweg-iota, line 29: I think you can use [const generics](https://blog.rust-lang.org/inside-rust/2021/09/06/Splitting-const-generics.html) here

vuongDang: Did not know about this feature, this is really cool!
This part is copied from previous GuardedVec and I have not planned any refactoring yet, but will definetely think about it when I do.

tensor-programming: Had the same thought myself when I was reviewing the code the other day and remembered that they had added the const  generics.  If I might rewrite this part upstream and commit it back to stronghold.

engine/new_runtime/src/buffer.rs
```diff
@@ -0,0 +1,370 @@
...
+};
+
+/// GuardedMemory is used when we want to store sensitive non encrypted data
+/// This shall always be short lived
+#[derive(Clone, Eq)]
+pub struct Buffer<T: Bytes> {
+    boxed : Boxed<T>, // the boxed type of current GuardedVec
+}
+
+pub struct Ref<'a, T: Bytes> {
```
felsweg-iota, line 20: nitpicking: wouldn't the `Ref` struct conflict with `core::cell::Ref`? Is there an advantage not implement `Deref` and `DerefMut`, but have a distinct type, expect for the lifetime?

felsweg-iota: nitpicking: wouldn't the `Ref` struct conflict with `core::cell::Ref`? Is there an advantage not implement `Deref` and `DerefMut`, but have a distinct type, expect for the lifetime?

engine/new_runtime/src/noncontiguous_memory.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Shards of memory which composes a non contiguous memory
+enum MemoryShard<T: Zeroize + Bytes> {
+    File(FileMemory),
+    Ram(RamMemory<T>)
+}
+
+// We set the maximum number of shards to 8, this is an arbitrary value that can be discussed
+const MAX_SHARDS: usize = 8;
+
+pub struct NonContiguousMemory<T: Zeroize + Bytes> {
```
felsweg-iota: `MAX_SHARDS` could also be implemented via const generics.

engine/new_runtime/src/types/zero.rs
```diff
@@ -0,0 +1,30 @@
...
+pub unsafe trait Zeroed: ContiguousBytes {
+    fn zero(&mut self) {
+        unsafe { sodium_memzero(self.as_mut_bytes().as_mut_ptr() as *mut _, self.as_bytes().len()) }
+    }
+
+    unsafe fn copy_and_zero(&mut self, other: &mut Self) {
+        assert!(other.size() >= self.size(), "other must be larger than self");
+
+        assert!(
+            (self.as_u8_ptr() < other.as_u8_ptr() && self.as_u8_ptr().add(self.size()) <= other.as_u8_ptr())
```
felsweg-iota, line 18: Maybe there is a way to reduce this complex boolean statement :)

felsweg-iota (commented):
Very nice work! I also like the new concept of split memory behavior types into `LockedMemory` and `ProtectedMemory`. Passing a config into `unlock(..)` to provide a key is also a good idea to even separate the underlying types even further. Really looking forward to how this will progress :)

engine/new_runtime/README.md
```diff
@@ -0,0 +1,75 @@
...
+Values in protected memory are stored in clear. Those values are accessible by getting a reference through `borrow()` and `borrow_mut()`.
+Since the values are stored in clear instances of `ProtectedMemory` should be as short-lived as possible.
+
+`ProtectedMemory` possesses `alloc()` and `dealloc` functions.
+
+Currently we have a single type `Buffer` implementing the `ProtectedMemory` trait.
+
+## `LockedMemory`
+Locked memory is used to store sensitive data for long period of time.
+On top of having the same protections as `ProtectedMemory`, `LockedMemory` values are never not stored in clear. This means that even when scanning memory, an attacker can't read directly the sensitive data from a dump.
```
elenaf9:
```suggestion
On top of having the same protections as `ProtectedMemory`, `LockedMemory` values are never stored in clear. This means that even when scanning memory, an attacker can't read directly the sensitive data from a dump.
```
typo?

elenaf9: > On top of having the same protections as `ProtectedMemory`, `LockedMemory` values (..)

Why do we need protections like Guard Areas, Canaries, flags against memory dumps etc. if the data is encrypted in memory anyway?

In my opinion, it would make sense to separate these two traits completely (and from how in understand your code, that also how is currently implemented). 
So `ProtectedMemory` is used to implement security measures for types that are otherwise stored plain in memory, like the `Key` for example. 
`LockedMemory` on the other hand is encrypted in memory and needs a key for each operation to decrypt it, but does not need the measures of `ProtectedMemory`. Locked memory would be used to protect and access the secrets that are stored in the vaults, right? And if we follow the old implementation, on access a value that is stored in a `LockedMemory` would be decrypted and wrapped into `ProtectedMemory` so that it can be used in the operation?

So I guess it's more about the wording in the README; the actual implementation aligns more or less with what I am describing.

elenaf9 (commented):
One first comment; I'll do a proper first review tomorrow.

engine/new_runtime/src/locked_memory.rs
```diff
@@ -0,0 +1,115 @@
...
+    ZeroedConfig(),
+    BufferConfig(usize),
+}
+
+// Different possible configuration for the memory
+// This is still in development
+// Size is an Option type because size is required for memory allocation but
+// not necessary for unlocking the memory
+#[derive(Debug)]
+pub enum LockedConfiguration<P: BoxProvider> {
```
elenaf9: This configuration type feels a bit weird, at least the way how it is currently used as parameter in the methods of `LockedMemory`. From what I understand, it is used to 
1. configure if file_memory is used or ram_memory
2. configure if the memory is encrypted or not.
3. Optionally pass the allocation size in case of alloc

- For Nr 1. we already decide that by **either** calling the methods of `EncryptedRam` **or** of `FileMemory`. Inside of these implementation the information is not useful anymore; on the contrary: it result in the error type `MemoryError::ConfigurationNotAllowed`, which in practice will never happen.
- Nr 2. could be added as a simple `Option<Key<_>>` parameter to the methods, not need for an enum here.
- Nr 3. can be just a parameter on `LockedMemory::alloc`. This way, it would be ensured on a type level that it is always present on `alloc`, but not needed for `dealloc`.
- For `ZeroedConfig` I currently don't see how/ when it is are relevant.

The above being said, I think we might need such a configuration enum on a higher level. So for example to configure what Memory type should be used we could have a config:
```rust 
pub enum MemoryType {
    File,
    Ram
    NonContiguousInRamAndFile,
}
```
and based on that call one of the 3 `LockedMemory` implementations.
But I would even then still advocate for having the key and the data-size as separate parameters.

engine/new_runtime/src/crypto_utils/crypto_box.rs
```diff
@@ -0,0 +1,154 @@
...
+
+/// A provider interface between the vault and a crypto box. See libsodium's [secretbox](https://libsodium.gitbook.io/doc/secret-key_cryptography/secretbox) for an example.
+pub trait BoxProvider: 'static + Sized + Ord + PartialOrd {
+    type Error: Debug;
+
+    /// defines the key length for the [`BoxProvider`].
+    fn box_key_len() -> usize;
+    /// defines the size of the Nonce combined with the Ad for the [`BoxProvider`].
+    fn box_overhead() -> usize;
+
```
elenaf9:
```suggestion
pub trait BoxProvider: 'static + Sized + Ord + PartialOrd {
    type Error: Debug;
    /// defines the key length for the [`BoxProvider`].
    const BOXED_KEY_LEN: usize;
    /// defines the size of the Nonce combined with the Ad for the [`BoxProvider`].
    const BOXED_OVERHEAD: usize;

```
I know this is old code, but couldn't we use [associated consts](https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants) for this?

engine/new_runtime/src/memories/encrypted_ram.rs
```diff
@@ -0,0 +1,206 @@
...
+use core::marker::PhantomData;
+use zeroize::{Zeroize};
+
+use serde::{
+    de::{Deserialize, Deserializer, SeqAccess, Visitor},
+    ser::{Serialize, Serializer},
+};
+
+/// GuardedMemory is used when we want to store sensitive non encrypted data
+/// This shall always be short lived
```
elenaf9: Why is this short lived? I would think it is used e.g. for storing the (encrypted) Record blob in memory, in which case it will be living in memory throughout the whole session. Since it is encrypted anyway, I would think that's not so much of a vulnerability?

engine/new_runtime/src/memories/encrypted_ram.rs
```diff
@@ -0,0 +1,206 @@
...
+
+// We currently only implement for u8 because our encryption functions only returns Vec[u8], therefore our cypher is Buffer<u8>
+impl<P: BoxProvider, const AD_SIZE: usize> LockedMemory<u8, P> for EncryptedRam<P, AD_SIZE> {
+    fn alloc(payload: &[u8], config: LockedConfiguration<P>) -> Result<Self, MemoryError> {
+        match config {
+            // For encrypted memory we don't store the key itself.
+            EncryptedRamConfig(key, size) => {
+                if size.is_none() {
+                    return Err(SizeNeededForAllocation);
+                }
```
elenaf9: See other comment; size can just be a parameter of `LockedMemory::alloc`, so we don't have to check if it is `None` and can save this error type here.

engine/new_runtime/src/memories/encrypted_ram.rs
```diff
@@ -0,0 +1,206 @@
...
+            let data = P::box_open(&key, &self.ad, &*self.cypher.borrow()).or(Err(DecryptionError))?;
+            if let EncryptedRamConfig(_, Some(size)) = self.config {
+                Buffer::alloc(&data, BufferConfig(size))
+            } else {
+                panic!("This should not happen if EncryptedRam has been allocated properly")
+            }
+        } else {
+            Err(ConfigurationNotAllowed)
+        }
+    }
```
elenaf9: I feel like the naming of the methods does not really match what they are doing. They actually don't change if/how the data is protected in memory. Instead `unlock` just simply borrows the data that is stored, and `lock` replaces the stored payload with a new payload. So if for example a user only reads a secret and never modifies it, it is not necessary to call `lock` (and unnecessarily re-allocate) because the payload didn't change. Same in `FileMemory`.

elenaf9 (commented):
Nice work! I very much like your abstraction for the security measures, and the different memory implementations.  
Couple of comments below. Looking forward to seeing the implementation of `LockedMemory` for non-contiguous memory.

engine/new_runtime/src/buffer.rs
vuongDang: It's necessary for `Boxed` type to have their own type for ref and refMut because there is a manual handling of the memory protections. In particular, when dropping a `Ref` or a `RefMut` you need to update the references counter and update the memory protections and I don't see how to do this without having distinct type.

engine/new_runtime/src/noncontiguous_memory.rs
vuongDang: I've removed the dynamic number of shards and just use 2 shards because it does not really improve security to have more than 2 shards

engine/new_runtime/README.md
vuongDang: This is a good remark. Right now all the memory types use guards and canaries and we can argue the necessity of such measures when you have encrypted data. The only argument I have right now is that canaries and guards help by protecting against data tempering through buffer overflow but not against privacy of this data. 
To be honnest I was thinking of choosing exactly how much security we implement depending on the performance we will get by using these new memory types.

engine/new_runtime/src/memories/encrypted_ram.rs
vuongDang: Ups, I think it's leftover copypasting, indeed `EncryptedRam` will be used to store data during the whole session

engine/new_runtime/src/memories/encrypted_ram.rs
vuongDang: Well that's a good point haha
Ok I will try to think of something else but feel free to give me some suggestions because my naming sense is terrible

engine/new_runtime/src/locked_memory.rs
vuongDang: I agree that the `Configuration` feels a bit clunky, I'm not 100% happy with this part. 

2. Yea that was the first idea but I was thinking that if we require other params in the future it would be cleaner to add it in the config rather than putting everything as an option in the function call
3. I like your idea for this one, I will do that

`ZeroedConfig` is used when zeroing the memory it's a bit like a default value.

elenaf9: > Yea that was the first idea but I was thinking that if we require other params in the future it would be cleaner to add it in the config rather than putting everything as an option in the function call

Good point. Yes if we add more properties then it might make sense. Maybe we can leave it as it is for now and once we have the full implementation and integration we can revive this discussion.

> `ZeroedConfig` is used when zeroing the memory it's a bit like a default value.

But do we need the config as a property in `EncryptedRam` / `FileMemory` at all? The only useful information it holds is whether the data is encrypted or not, but that could just be set as a bool, no? Because we replace the `Key` itself with a random value anyway, and also the `size` information is not used afterwards anymore. If we remove it, then also the "default" `ZeroedConfig` wouldn't be needed anymore.

engine/new_runtime/src/memories/encrypted_ram.rs
elenaf9: Well we could just go with simple `unlock` -> `borrow` and `lock` -> `write`. Or we merge it to the one method `exec_on_unlock` that you listed as an idea, which takes a closure `Fn(Buffer<u8>) -> Option<Buffer<u8>>`. If the return value is `Some(data)` then we do the re-allocation, otherwise the stored payload does not change. But as you already mentioned as a _Con_ there is that it could get difficult if we have multiple steps of encryption.

engine/new_runtime/src/crypto_utils/crypto_box.rs
tensor-programming: Yeah you could.  They weren't an option when I wrote this.

engine/new_runtime/README.md
```diff
@@ -0,0 +1,78 @@
+# Stronghold new runtime 
+
+This crate provides multiple types of secure memories which implementation is abstracted behind two interfaces `ProtectedMemory` and `LockedMemory`.
```
felsweg-iota:
```suggestion
This crate provides multiple types of secure memory implementations, that are abstracted behind two interfaces `ProtectedMemory` and `LockedMemory`. 
```

engine/new_runtime/README.md
```diff
@@ -0,0 +1,78 @@
+# Stronghold new runtime 
+
+This crate provides multiple types of secure memories which implementation is abstracted behind two interfaces `ProtectedMemory` and `LockedMemory`. 
+These interfaces are both present in the file _src/locked_memory.rs_
```
felsweg-iota:
```suggestion
The interfaces are both present in the file _src/locked_memory.rs_
```

engine/new_runtime/README.md
felsweg-iota: I am disagreeing to the point where encrypted memory shouldn't have any protection, because is encrypted. The major point of the new runtime provision is to keep sensitive data as secure as possible. If encrypted memory can also not be dumped, we counter attacks to retrieve memory from minimal timed exposure as well. This should be top prio, performance is second.

engine/new_runtime/README.md
```diff
@@ -0,0 +1,78 @@
...
+  - split in both memory and file system
+- Both at the same time
+
+`LockedMemory` possesses `alloc()`, `dealloc`, `lock()` and `unlock()`.
+To use a `LockedMemory` one needs to unlock it before.
+
+## Ideas 
+- Instead of having to unlock then lock `LockedMemory` every time, we could have a single function `exec_on_unlock()` to which we provide a closure manipulating "unlocked" memory. This is closer to the current `GuardedVec`.
+  + Pros: the unlocking is done automatically at the end of the closure
+  + Cons: we have some data which is encrypted and those keys are also `LockedMemory`. Therefore unlocking multiple layers of `LockedMemory` would force us to have nested closures which could be ugly.
```
felsweg-iota: I would suggest to focus on the intended operation. If some sensitive data requires a key to be decrypted as well, the API should reflect that, or make it opaque providing the decrypted part as closure argument. This API could reside on a higher level abstraction.

engine/new_runtime/src/locked_memory.rs
```diff
@@ -0,0 +1,123 @@
...
+    // Non contiguous non encrypted memory in ram and disk
+    NCRamAndFileConfig(Option<usize>),
+
+    // Non contiguous non encrypted memory in ram
+    NCRamConfig(Option<usize>),
+}
+
+impl<P: BoxProvider> LockedConfiguration<P> {
+    // Check that variants type are the same
+    pub fn is_eq_config_type(&self, other: &Self) -> bool {
```
felsweg-iota: you can use ` std::mem::discriminant(...)` to uniquely compare the variants inside the `LockedConfiguration` enum, while disregarding their value.

engine/new_runtime/src/locked_memory.rs
felsweg-iota: I wonder, if we shouldn't make a distinction of the memory type concerning:

- the location
- the kind of memory
- the encryption state

```rust
enum MemoryLocation {
   /// Data shall reside in RAM
   Memory,
   
   /// Data shall reside in file-system
   File,

   /// This would be the combination of both, as providing it 
   /// somewhere else would require some semantic on the 
   /// provided ordering
   FileRam
}

enum MemoryKind {
    /// A continguous stream of bytes
    Continguous,
    
    /// Noncontiguous stream of bytes
    NonContinguous
}

enum MemorySecurity<P>  where P :Provider{
    /// Encrypted with key
    Encrypted(Key<P>),

    /// Plain memory type 
    Plain
}

/// all three types could be part of a configuration type
struct LockedMemoryConfiguration<P> where P : Provider {
    location : MemoryLocation,
    kind: MemoryKind, 
    security : MemorySecurity<P>
}
```
The implementation could return sensible defaults / complex setups

```rust
impl<P> LockedMemoryConfiguration {
     
     /// Returns a [`LockedMemoryConfiguration`]
     pub fn encrypted_non_contiguous_memory_file(&self, key : Key<P>) -> Self {
         Self {
               location : MemoryLocation::Both,
               kind : MemoryKind::NonContinguous,
               security : MemorySecurity::Encrypted(key)
         }
     }

     /// more can be added, if needed
}
```

The advantage would be to have a distinction between the different types, while keeping them together inside a configuration struct, that provides sensible default constructors.

felsweg-iota: I would also suggest to either put the `LockedMemory` and `ProtectedMemory` traits either in separate modules, or rename the module accordingly.

engine/new_runtime/src/locked_memory.rs
```diff
@@ -0,0 +1,123 @@
...
+            (NCRamConfig(s1), NCRamConfig(s2)) => s1 == s2,
+            (_, _) => false
+        }
+    }
+}
+
+impl<P: BoxProvider> Eq for LockedConfiguration<P> {}
+
+/// Memory storage with default protections to store sensitive data
+pub trait ProtectedMemory<T: Bytes>: Debug + Sized + Zeroize + Drop {
```
felsweg-iota: I know that a specific implementation of `Drop` is required for types, but `Drop` will always be automatically provided. I wonder, if we could instead provide a more explicit trait, that explicitly calls `drop(...)` on a type.

engine/new_runtime/src/memories/buffer.rs
```diff
@@ -0,0 +1,370 @@
...
+        let guard = Buffer::<u8>::zero(1);
+
+        let moved = guard;
+
+        assert_eq!(*moved.borrow(), [0]);
+
+        let guard = Buffer::<u8>::random(8);
+
+        let borrow = guard.borrow();
+        let clone = borrow.clone();
```
felsweg-iota, line 328: just saw it: the `Boxed` type provides a Clone.

engine/new_runtime/src/memories/encrypted_ram.rs
```diff
@@ -0,0 +1,209 @@
...
+
+// Additional data used as nonce
+const AD_SIZE: usize = 32;
+
+/// GuardedMemory is used when we want to store sensitive non encrypted data
+/// This shall always be short lived
+pub struct EncryptedRam<P: BoxProvider> {
+    // Visibility within crate is for testing purpose
+    // We currently only implement for u8 because our encryption functions only returns Vec[u8], therefore our cypher is Buffer<u8>
+    cypher: Buffer<u8>,
```
felsweg-iota:
```suggestion
    cipher: Buffer<u8>,
```

engine/new_runtime/src/memories/encrypted_ram.rs
```diff
@@ -0,0 +1,209 @@
...
+    // Configuration, we should only allow EncryptedRamConfig for this struct
+    // We do not store the key in the struct config, just random data for security
+    config: LockedConfiguration<P>,
+    // Associated data, we will use it as a nonce with random value
+    ad: [u8; AD_SIZE],
+}
+
+// We currently only implement for u8 because our encryption functions only returns Vec[u8], therefore our cypher is Buffer<u8>
```
felsweg-iota:
```suggestion
// We currently only implement for u8 because our encryption functions only returns Vec[u8], therefore our cipher is Buffer<u8>
```

engine/new_runtime/src/memories/encrypted_ram.rs
felsweg-iota: Agree, `alloc` should have a size parameter, on the other hand we can think of inferring the allocation size by  `std::mem::sizeof<T>()`.

engine/new_runtime/src/memories/file_memory.rs
```diff
@@ -0,0 +1,220 @@
...
+        let fname: String = FileMemory::<P>::random_fname();
+        let mut file = File::create(&fname)?;
+        file.write_all(payload.as_bytes())?;
+        Ok(fname)
+    }
+
+    fn clear_and_delete_file(&self) -> Result<(), std::io::Error> {
+        let mut file = File::create(&self.fname)?;
+        // Zeroes out the file
+        file.write_all(&vec![0; self.get_data_size()])?;
```
felsweg-iota: I wonder, if zeroing out a file works for all types of file-systems. Solid state drives have a special caching mechanism, where writes are not directly conducted by buffered in between. Earlier hard-disks could be securely wiped with multiple overwrites of random data. maybe that's just overkill since the shards inside file system are encrypted.

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+
+
+static IMPOSSIBLE_CASE: &'static str = "NonContiguousMemory: this case should not happen if allocated properly";
+
+// Currently we only support data of 32 bytes in noncontiguous memory
+const NC_DATA_SIZE: usize = 32;
+
+// NONCONTIGUOUS MEMORY
+/// Shards of memory which composes a non contiguous memory
+enum MemoryShard<T: Zeroize + Bytes, P: BoxProvider> {
```
felsweg-iota: Minor: Avoid repeating a suffix in variants, if the enum itself makes the intend clear.

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+    // EncryptedFileShard(FileMemory<P>),
+    // EncryptedRamShard(EncryptedRam<P>),
+    ZeroedShard(),
+    FileShard(FileMemory<P>),
+    RamShard(Buffer<T>),
+}
+use MemoryShard::*;
```
felsweg-iota, line 43: Better use it directly in calling functions

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+use crate::types::Bytes;
+use crypto::hashes::sha;
+use core::fmt::{self, Debug, Formatter};
+use zeroize::Zeroize;
+
+
+static IMPOSSIBLE_CASE: &'static str = "NonContiguousMemory: this case should not happen if allocated properly";
+
+// Currently we only support data of 32 bytes in noncontiguous memory
+const NC_DATA_SIZE: usize = 32;
```
felsweg-iota: Does this restriction have a reason, or is it just easier to partition / reassemble 32 byte chunks? I wonder, if this could be more flexible, but it's not a requirement. A possible extension could be a String of non-contiguous memory.

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+        match &self.shard1 {
+            RamShard(buf) => buf,
+            _ => panic!("{}", IMPOSSIBLE_CASE),
+        }
+    }
+
+
+    // Refresh the shards to increase security, may be called every _n_ seconds or
+    // punctually
+    fn refresh(self) -> Result<Self, MemoryError> {
```
felsweg-iota, line 122: I think, that's a good solution for now to not directly provide the means to "automatically" flicker the memory parts, but lay it out to a higher level abstraction.

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+                    .expect("Failed to unlock file memory");
+                let new_data2 = xor(&buf.borrow(), &hash_of_old_shard1);
+                let new_data2 = xor(&new_data2, &hash_of_new_shard1);
+                let new_fm = FileMemory::alloc(&new_data2,
+                                               FileConfig(Some(NC_DATA_SIZE)))?;
+                new_shard2 = FileShard(new_fm);
+                // new_shard2 = FileShard(fm.lock(new_data2, FileConfig(Some(NC_DATA_SIZE)))?);
+                new_config = NCRamAndFileConfig(Some(NC_DATA_SIZE));
+            },
+            _ => panic!("{}", IMPOSSIBLE_CASE)
```
felsweg-iota: Maybe better return an error indicating a non-handled case of an enum variant.

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+        }
+    }
+}
+
+impl<P: BoxProvider> Zeroize for NonContiguousMemory<P>
+{
+    fn zeroize(&mut self) {
+        self.shard1.zeroize();
+        self.shard2.zeroize();
+        self.config = LockedConfiguration::ZeroedConfig();
```
felsweg-iota: Minor: A zero argument variant shouldn't have an empty tuple as data. See also the discussion above concerning a ZeroedConfig as default

felsweg-iota (commented):
Good progress so far! Also like your implementation of the non-contiguous types. I wonder, if some of the proposed changes to the configuration to the various memory types would ease up the implementation.

engine/new_runtime/src/memories/file_memory.rs
vuongDang: Did not know about it, I will add it to my list of things to watch out for because I don't have any obvious solution for this currently

engine/new_runtime/src/locked_memory.rs
vuongDang: I wanted to use the trait `ZeroizeOnDrop` but could not make it work which automatically calls `zeroize()` when dropping. 
However I could not make it work (don't remember why) and pushed it to later work.

engine/new_runtime/src/memories/noncontiguous_memory.rs
vuongDang: The only hard requirement is that NC memory should have the same size of the output of the hashing algorithm used. 
In our case NC memory will be used for keys which are 32 bytes in the current implementation. 
I agree that it should and could be more flexible.

engine/new_runtime/src/memories/noncontiguous_memory.rs
vuongDang: Isn't it better to panic if we reach a case that should never happen and is present just because Rust requires exhaustive pattern matching?

elenaf9: How about using the rust idiomatic `unreachable!("Config was already checked.")`? Does the same as `panic`, but emphasizes that this case will **never** happen because we already checked it above.

vuongDang: Sounds perfect for this use case, thanks

felsweg-iota: Depends on what we want to achieve. If we extend `LockedConfiguration`, but do not provide a matching implementation for new types, I would expect an error. Since `refresh` doesn't accept a config type from the outside, this is a logical follow up error. I would also suggest to avoid `unreachable!()` in that case, because this case should be handled earlier. I do understand the desired implication. Maybe `unreachable!()` can be used earlier, if we intent to "panic with reason" on this crucial part.

elenaf9: @felsweg-iota before this match block, in the same method, it is already checked if the config is one of the two valid ones, and otherwise an error is returned:
https://github.com/iotaledger/stronghold.rs/blob/43cde6b3be50a4f0b433e4551b3d17648e65f617/engine/new_runtime/src/memories/noncontiguous_memory.rs#L132
So it is already doing exactly what you are proposing. Am I correct assuming that you missed this line?

felsweg-iota: @elenaf9 no. I see no reason to introduce `unreachable!()` here, where some returned error could indicate a failure to allocate memory. I would contain `panic` etc. to more sensitive regions. While this code indicates prior-allocation, but config checking, an error would be appropriate.

elenaf9: > I see no reason to introduce `unreachable!()` here

The reason is that this code is, as a matter of fact, unreachable 😄.

elenaf9: No actually, you are right @felsweg-iota. I did not read through the code carefully enough and missed that it is matching `self.shard2` here, sorry. I thought it was matching on `self.config`.
In this case I agree with you that we should return an error. Or we could extend `NonContiguousMemory::check_config` to also check that `self.shard1` and `self.shard2` are valid.
Or do we even need `MemoryShard::ZeroedShard` at all? It is never used, not even in `zeroize`.

Issue #278, opened 2021-11-04
State Synchronization Across Multiple Stronghold Instances


felsweg-iota:
## Description

Local and remote Stronghold state synchronization via snapshots. 

## Motivation

The synchronization between multiple instances of Stronghold may occur, when the same application is available on multiple platforms storing secret information but is not used at the same time. 

Another use case is remote synchronization of state across multiple instances of Stronghold. This would involve a more elaborate protocol to securely transfer secrets across insecure environments.

## Requirements

Write a list of what you want this feature to do.

1. Synchronize locally: full and partial
2. Synchronize remote: secure protocol needed

## Are you planning to do it yourself in a pull request?

Yes

PR #317, opened 2022-02-09
Feat(client): state synchronization of snapshots and clients

elenaf9:
# Description of change

This PR implements a synchronization feature that allows to fully or partially merge the vaults between two structures.
For this, it introduces a new trait `MergeLayer` for merging two instances, and implements it for the Client-Layer (= sync of two client states) and Snapshot-Layer (= sync of two snapshot states).
The steps for a full synchronizing of an instance A with an instance B (i.e. A would like to import all records from B that it does not have yet) are:
1. get the full hierarchy of all ids stored B
2. compare this hierarchy at A with the local hierarchy to get the diff
3. export all entries from B that are in this diff
4. import entries at A

Since the keys usually differ between two instances, the old keystore(s) should be provided on import to decrypt each record with the old key and re-encrypt it with the local one.
In case of a remote sync, the exported entries are imported to an empty snapshot that internally populates a new keystore with temporary keys. At the remote the entries are then exported from received state and imported to the local state.

The described fundamental steps can be enhanced with the optional mapper if A and B have different Ids for the same client / vault / record. This mapper maps one full path (`VaultId` ++ `RecordId` in case of client-sync, `ClientId` ++ `VaultId` ++ `RecordId` in case of snapshot-sync) to another, optionally returning a `None` in case of a partial sync where only some of the records should be imported.
For handling conflicts (a client / vault / records exists both, at A and B, but differs in the content) a merge policy is required from the user, which specifies whether to keep the old values, replace it with the new, or recursively merge the entries.

_**Note:** For remote sync, certain functionalities for loading and persisting client states are required (`Stronghold::fill_snapshot_state` and `Stronghold::load_state`). These were still in discussion in https://github.com/iotaledger/stronghold.rs/pull/296#issuecomment-982283276, and will be solved after the upcoming refactoring once we implement the RFC of #308. However, I had to add them here already in order to integrate remote sync into the interface. @tensor-programming since you were against adding them, I would propose to finish this PR, but not merge it before the refactoring. Once we have the refactoring done, I can change the sync-feature to use the new, more secure architecture._

_(Another) Small Note: The priority in the implementation is to be fully flexible for the different use cases, but introduce as little overhead as possible. Therefore it attempts to only do a re-encryption if it is really necessary, and drop records according to the merge policies as early as possible if they won't be imported anyway. With this the `MergeLayer::import_entries` implementation for the two layer turned out to have a lot of duplicated code for the different cases (Mapper: yes/no; Old-Keystore: yes/no). I did not find a cleaner way to implement this, but I would be grateful for any ideas on how two improve the code quality there._

TODOs:
 - [x] Remove unwraps, add proper error handling
 - [x] Remote sync
    - [x] implement remote sync 
    - [x] fix serialize for HashMap with `client::utils::ids::Id` as key:
    > Serialization can fail (..) if `T` contains a map with non-string key 
 - [x] Integration into interface
 - [x] Encrypt snapshot key on remote sync  

Issue #278. Alternative implementation to #292, minus the policy engine.

## Type of change

- unit tests `sync::test::{test_get_hierarchy, test_export_with_mapping, test_merge_policy}`.
- integration tests: `tests::synchronization_tests::{sync_clients, remote_sync}`

## Change checklist


client/src/state/snapshot.rs
```diff
@@ -76,6 +80,101 @@ impl Snapshot {
...
+        self.state.0.insert(cid0, state0);
+        self.state.0.insert(cid1, state1);
+    }
+
+    /// Sync the local state with another snapshot, which imports the entries from `other` to `local`.
+    pub fn sync_with_snapshot(
+        &mut self,
+        other: &mut SnapshotState,
+        mapper: Option<Mapper<(ClientId, VaultId, RecordId)>>,
+        merge_policy: SelectOrMerge<SelectOrMerge<SelectOne>>,
```
felsweg-iota: hm could the provided `merge_policy` type easier? It feels odd to have `SelectOrMerge<SelectOrMerge<_>>`, but rather a simple `SelectOrMerge`.

client/src/state/snapshot.rs
```diff
@@ -76,6 +80,101 @@ impl Snapshot {
...
+    }
+
+    /// Export the entries specified in the diff to a blank snapshot state.
+    /// This re-encrypts the entries with new vault-keys that are inserted in the new
+    /// snapshot state alongside with the exported records.
+    ///
+    /// Deserialize, encrypt and compress the new state to a bytestring.
+    pub fn export_to_serialized_state(
+        &mut self,
+        diff: <sync::SnapshotState as MergeLayer>::Hierarchy,
```
felsweg-iota: Better extract the functions provided for `Snapshot` into an own trait, with respective bounds to provided types. `Hierarchy` can here be provided as the to be synced type.

client/src/state/snapshot.rs
```diff
@@ -130,13 +130,13 @@ impl Snapshot {
     /// Deserialize, encrypt and compress the new state to a bytestring.
     pub fn export_to_serialized_state(
         &mut self,
-        diff: <sync::SnapshotState as MergeLayer>::Hierarchy,
+        diff: <SyncSnapshotState as MergeLayer>::Hierarchy,
```
felsweg-iota: Better pass it as generic: `<SyncSnapshotState as MergeLayer>::Hierarchy` could some function generic `H`

client/src/state/snapshot.rs
```diff
@@ -159,28 +159,31 @@ impl Snapshot {
...
-    pub fn get_hierarchy(&mut self) -> <sync::SnapshotState as MergeLayer>::Hierarchy {
-        sync::SnapshotState::from(&mut self.state).get_hierarchy()
+    pub fn get_hierarchy(&mut self) -> <SyncSnapshotState as MergeLayer>::Hierarchy {
+        SyncSnapshotState::new(&mut self.state).get_hierarchy()
     }
 
     /// Calculate diff between local hierarchy and the given one.
     pub fn get_diff(
         &mut self,
-        other: <sync::SnapshotState as MergeLayer>::Hierarchy,
```
felsweg-iota: See above

felsweg-iota (commented):
Massive synchronization code here :+1: You try to solve a lot in a few functions, so I would recommend to reduce the complexity on certain parts eg. `import_entries`. I am also not quite sure why you partially employ a recursive enum to define a merging policy:

> ```rust
>#[derive(Debug, Clone, Copy)]
>pub enum SelectOne {
>    KeepOld,
>    Replace,
>}
>
>/// Policy for selecting between two instances (e.g. client states or vaults) with the same id.
>#[derive(Debug, Clone, Copy)]
>pub enum SelectOrMerge<T> {
>    KeepOld,
>    Replace,
>    /// Recursively merge inner entries.
>    Merge(T),
> }
> ```

If I assume correctly, you are trying to directly represent the hierarchy of the underlying data structure as policy. Why not split the operations up, and supply the respective function calls with a vector of policies, if any? This would reduce the complexity, but maybe I'm missing a point here.

client/src/state/snapshot.rs
elenaf9: That's not possible here, because below in line 136 `export_entries` expects this exact type and would not work with a generic `H`:
```rust
let exported = SyncSnapshotState::new(&mut self.state).export_entries(Some(diff));
```

client/src/state/snapshot.rs
elenaf9: If we would only have it on one layer it would be better, I agree. But we have this on three layers:
- Vault layer: only `SelectOne` (select-one between two records)
- Client layer: `SelectOrMerge<SelectOne>` (Select-or-merge between two vaults, select-one between two records)
- Snpashot layer: SelectOrMerge<`SelectOrMerge<SelectOne>>` (Select-or-merge between two clients, Select-or-merge between two vaults, select-one between two records)
If we translate this into one enum each, Imo the recursion in this would be less transparent, and instead we would end up with larger enums (4 variants for client layer, 6 variants on snapshot layer) that are a bit redundant. Furthermore, because the `<SnapshotState as MergeLayer>` implementation internally recursively calls `<ClientState as MergeLayer>`, we would also then have to map between the two types.
That  being said, I don't have a strong opinion on this, so if you insist I can change it to a single enum each.

client/src/state/snapshot.rs
felsweg-iota: hm not quite sure to use an inner type as an outer type :)

elenaf9: Good point, I agree that it is weird to expose to associated type of an inner trait like this. I changed it in 4af3773 to just use the actual types directly.

client/src/interface.rs
```diff
@@ -806,4 +912,71 @@ impl Stronghold {
     async fn network_actor(&self) -> StrongholdResult<Addr<NetworkActor>> {
         self.registry.send(GetNetwork).await?.ok_or(ActorError::TargetNotFound)
     }
+
+    /// Sync the local stronghold with a remote one.
+    pub async fn sync_with(
+        &self,
+        peer: PeerId,
+        merge_policy: SelectOrMerge<SelectOrMerge<SelectOne>>,
```
felsweg-iota, line 937: it's really hard to decipher this type `SelectOrMerge<SelectOrMerge<SelectOne>>`. Since it's a policy, I would expect some kind of configuration type; maybe a solution to this could be some configuration variants implemented as functions, that returns a specific strategy.

client/src/sync.rs
```diff
@@ -0,0 +1,831 @@
...
+
+    fn get_diff(
+        &mut self,
+        other: Self::Hierarchy,
+        mapper: Option<&Mapper<Self::Path>>,
+        merge_policy: &Self::MergePolicy,
+    ) -> Result<Self::Hierarchy, VaultError> {
+        let mut diff = HashMap::<VaultId, Vec<_>>::new();
+        for (vid0, records) in other {
+            for (rid0, bid) in records {
```
felsweg-iota, line 166: I wonder, if `BlobId`s can be left out of the synchronization part. I would assume, that the relevant part is the id of the record, since the contents might be outdated, and should be synchronized, otherwise this could be part of a `MergePolicy`, than source is being retained, target being update etc...

felsweg-iota: Ok. Got it. you later differentiate according to the `SelectOrMerge` named policy. Maybe split it into multiple functions each having distinct functionality

client/src/sync.rs
```diff
@@ -0,0 +1,831 @@
...
+    fn import_entries(
+        &mut self,
+        exported: Self::Exported,
+        merge_policy: &Self::MergePolicy,
+        mapper: Option<&Mapper<Self::Path>>,
+        old_key_provider: Option<&Self::KeyProvider>,
+    ) -> Result<(), VaultError> {
+        let mut mapped: HashMap<VaultId, Vec<_>> = HashMap::new();
+        match mapper {
+            Some(mapper) => {
```
felsweg-iota, line 217: See above. Better split functionality to have distinct functions.

felsweg-iota (commented):
Very good approach to generalize the synchronization between local snapshots ( currently bound to clients ), and remote state. Please take a look into the (few) comments inline and see, if you can simplify the merge policy configuration. Maybe it's just an enum with structured variants, but that's just an idea. Also good job on remote synchronization. I think your solution with a DH based approach on top of p2p is excellent.

client/src/interface.rs
elenaf9: > Since it's a policy, I would expect some kind of configuration type; maybe a solution to this could be some configuration variants implemented as functions, that returns a specific strategy.

I am not sure if I understand what you mean. Could you draft an example for such a function?

felsweg-iota: Sure. I assume the problem to be solved with synchronization is the currently implied type-hierarchy by Stronghold. You chose an approach to reflect this hierarchy over generic types, which are effectively restricted to `SelectOrMerge` or `SelectOne`, but can potentially anything else. While this is being used to unpack the inner types and match on them, I wonder if a more direct approach, maybe a bit more verbose, would make sense. I'm drafting out an idea, which may not be complete but could give a direction:

The assumption is that synchronization is always an operation on a pair A -> B, where A _wants to synchronize_ with B.

```rust
/// A specific merge policy type
enum MergePolicy {
    /// This would keep the old entries in A, adding only newer / non-existing ones from B
    Retain,
    
    /// This would update entries in A to B
    Update,
}
``` 

The `MergePolicy` could reside inside a `MergeConfiguration`. 
```rust
struct MergeConfiguration {
     vaults: MergePolicy,
     records: MergePolicy,
     store: Option<MergePolicy>
}
```

Now a `MergeConfiguration` could be constructed via creation functions: 

```rust
impl MergeConfigurtion {
    
    /// the function name could definitely be better.
    pub fn update_records_keep_vaults_discard_store() -> Self {
        Self { 
             vaults: MergePolicy::Retain,
             records: MergePolicy::Update,
             store: None
        }
    }
}
```
This way we could bundle the intended synchronization policies for each respective parts while being verbose and clear what part is affected. This would go against your more generic approach, but I would assume that switching the config / policy types would make it easier to reason about.

elenaf9:
Closing this PR in favor of #334, which implements a simpler version of the synchronization feature. This PR may still be used as reference when adding remote synchronization.

PR #319, opened 2022-02-21
Refactor(client): simplify procedure API

elenaf9:
# Description of change

Simplify the procedure API added in #258.
The main objective of this PR is to:
1. Reduce the overall number of different types and traits related to the procedure API
2. Remove the derive macro to ease future maintaining of the procedure API
3. Improve convenience for executing only a single procedure
4. Enable that all operations on the vault can be done through procedures (e.g. write to vault, revoke data, garbage collect data)
5. Remove unnecessary complexity of procedure input and output (fixed vs dynamic input; temporary vs permanent output)

Internally, most logic and convenience added with #258 is preserved:
- New procedures can easily be added by implementing on of the 3 traits: `GenerateSecret`, `DeriveSecret` or `UseSecret`
- Multiple procedures can be executed in a row (especially relevant in a remote setting); the collected output is returned
- All new cryptographic procedures are still supported (minus the `Hash` procedure)

From a user perspective, the procedure API is now much more aligned with the "old" implementation (before #258):
- Removed concept of temporary output and the necessity to explicitly persist secret and non-secret output with the `PersistOutput` and `PersistSecret` traits:
  - if a secret is only temporary needed, the user can manually revoke it again with the new `RevokeData` procedure
  - irrelevant non-secret output can just be ignored
-  remove `new` constructor for each procedure, instead make fields public 
- add `Output` type for each procedure. If only a single procedure is executed, the return type is directly converted to this output type (e.g. `()` in case of WriteVault, or `ChainCode` in case of `Slip10Derive`)
- if multiple procedure are executed: expect a `Vec` of the procedures, return a `Vec` of the results of each procedure in the same order

**To be discussed:**

The feature of "dynamic" non-secret input is removed: it is not possible anymore that one procedure takes the non-secret output of another procedure in the same execution chain as input. I could not think of a case where this is really needed in practice, and removing this feature drastically reduced the complexity of the API and enabled most of the other simplification. In case that anyone thinks this feature is needed, I am happy to re-think a cleaner way of re-adding it.

//CC @abdulmth.

## Type of change

Tested with existing tests.

## Change checklist


client/src/interface.rs
```diff
@@ -33,7 +33,7 @@ use p2p::{identity::Keypair, DialErr, InitKeypair, ListenErr, ListenRelayErr, Ou
 
 use actix::prelude::*;
 use serde::{Deserialize, Serialize};
-use std::{path::PathBuf, time::Duration};
+use std::{convert::TryInto, path::PathBuf, time::Duration};
```
felsweg-iota, line 36:
```suggestion
use std::{path::PathBuf, time::Duration};
```
`TryInto` is already imported by default

client/src/procedures/primitives.rs
```diff
@@ -55,108 +43,204 @@ pub enum PrimitiveProcedure {
...
-enum_from_inner!(PrimitiveProcedure::Pbkdf2Hmac from Pbkdf2Hmac);
-enum_from_inner!(PrimitiveProcedure::AeadEncrypt from AeadEncrypt);
-enum_from_inner!(PrimitiveProcedure::AeadDecrypt from AeadDecrypt);
-
-// ==========================
-// Helper Procedure
-// ==========================
+/// Garbage collects any revokes in a Vault based on the given `vault_path`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct GarbageCollect {
```
felsweg-iota, line 189: I wonder, if `GarbageCollect` (GC) should remain as a single procedure, or if garbage collection could be run separately. The GC was provided as an atomic operation, but this could be part of a complete procedure call to automatically clean up unused locations.

client/src/procedures/types.rs
```diff
@@ -373,297 +50,179 @@ pub struct Products<T> {
...
-    /// [`OutputKey`] with which the procedure's output is associated.
-    /// This key can be used to specify the input of subsequent procedures in a procedure chain.
-    ///
-    /// If [`PersistOutput::store_output`] was not set, this key is random, and the output will not be
-    /// persisted in the [`CollectedOutput`] that is returned to the user.
-    fn output_key(&self) -> OutputKey;
-
-    /// Write the procedure's non-secret output into the [`CollectedOutput`] that is returned to the user.
-    fn store_output(self, key: OutputKey) -> Self;
+impl From<ProcedureOutput> for () {
```
felsweg-iota, line 144: hm I guess you have to provide the impl for an empty tuple, since the trait `Procedure` requires `type Output: TryFrom<ProcedureOutput>;` Other option would be to reduce the bound only to `ProcedureOutput`, call `into` would be done anyways and be present for all types, that are allowed as input.

felsweg-iota (approved):
Good job simplifying the procedures API. The interface is more clear, also the separation in deriving, generating and using procedures makes sense. With the upcoming runtime changes, there would still be some changes involved. `GuardedVec` will be replaced; I wonder if the different memory types should be opaque ( not visible to the user ) and just operate on the provided types, or if certain procedures could be specialized in terms of their usage eg. The non-contiguous mixed-memory types.

client/src/procedures/primitives.rs
elenaf9: > but this could be part of a complete procedure call to automatically clean up unused locations

@felsweg-iota  what do you mean by  "unused locations"?

client/src/procedures/types.rs
elenaf9: > Other option would be to reduce the bound only to `ProcedureOutput`

Do I understand you correctly that you propose that all Procedures should just have the return type `ProcedureOutput`?

The reason why we have the `Procedure::Output` associated type is so that the user can then safely convert it to the actual procedures output type. So if a Procedure has return type `String`, the value is intermediately converted to `ProcedureOutput`, but the user can then later convert it back to `String` and rely on the fact that this won't fail.
This is also done in `Stronghold::exec_proc`, where it enables that the return value of the method is the one of `Procedure::Output`. So that `stronghold.runtime_exec(WriteVault {..}) ` has return value `()`.
https://github.com/iotaledger/stronghold.rs/blob/2fc146afd8862bf5023f9f6e9e9b0828610f7f81/client/src/interface.rs#L294-L301

elenaf9:
Thank you for the review @felsweg-iota.

> With the upcoming runtime changes, there would still be some changes involved. GuardedVec will be replaced; I wonder if the different memory types should be opaque ( not visible to the user ) and just operate on the provided types, or if certain procedures could be specialized in terms of their usage eg. The non-contiguous mixed-memory types.

You are definitely right that parts of the procedure API will need to be adapted once the runtime changes are done. That being said, I think this is something we should tackle once the new runtime is actually finished and that this PR should not depend on it. I'd rather have multiple iterations on this, than to make any assumptions right now about the runtime while it is still in early development.
Given that this PR just cleans up a lot of complexity that was introduced in #258, I think it is already a improvement as it is. Especially if the identity team (cc @abdulmth) are currently updating their stronghold integration to the latest dev.

client/src/procedures/primitives.rs
felsweg-iota: sorry, wrong wording in this case. I am referring to locations that have been soft-deleted.

elenaf9: @felsweg-iota what do you mean by "soft-deleted"? 😅 Could you give an example?

client/src/procedures/types.rs
felsweg-iota: Makes sense, since output types are required to implement a conversion.

client/src/procedures/primitives.rs
felsweg-iota: "Soft-Deletion" is a term used in databases to indicate a record to be deleted, while effectively it is not. A record (inside a db table) gets a marker field, where the time of deletion is recorded, the record is thus revoked. At a later time, the record gets effectively deleted from the table.

Having said that, Stronghold also has an atomic pair of `revoke` / `garbage collect` operations. `delete_data` (in the present interface) does exactly that. A boolean value indicates either just `revokation` of data, or direct deletion.

elenaf9: So if I understand you correctly, you are proposing that after the procedure call we automatically run garbage-collect on the vaults where data has been revoked during the procedure call?

felsweg-iota: yes

elenaf9: Okay I think that's a good idea

> Having said that, Stronghold also has an atomic pair of revoke / garbage collect operations. delete_data (in the present interface) does exactly that. A boolean value indicates either just revokation of data, or direct deletion.

The new `RevokeData` procedure has this flag as well; if `should_gc` it garbage collects the vault from which the data was deleted. 

What I am wondering is whether we should just change `RevokeData` to always directly delete the record. 
I don't really understand the benefit in the flow of:
1. revoke-data: creating a `RevokeTransaction` for a record 
2. garbage-collect: iterate through all records from a vault and remove the records with such a `RevokeTransaction`

rather than just removing the record from the `HashMap` of records. 
@tensor-programming could you expand a bit on what the initial motivation was behind this? Do you see a reason not to just remove the record completely when revoking it?

client/src/interface.rs
elenaf9: `TryInto` is only in the prelude starting in Rust edition 2021. We are currently still using edition 2018: https://github.com/iotaledger/stronghold.rs/blob/312742c212c140adb0700e40d9af63af86da185e/client/Cargo.toml#L8

I think we should upgrade to edition 2021, but imo that should be done in a separate issue/ PR.

PR #320, opened 2022-02-24
dev-refactor merge 

felsweg-iota:
# Description of change

This is the branch for upcoming features. No explicit change, but to keep `dev-refactor` mergable with `dev`. Merge this PR into `dev`, when all other changes are completed. 

## Type of change

- [x] Documentation Fix

All dependent branches and features have thoroughly been tested by unit and integration tests. 

## Change checklist


.github/workflows/covector-version-or-publish.yml
```diff
@@ -10,6 +10,7 @@ on:
   push:
     branches:
       - dev
+      - dev-refactor
```
elenaf9, line 13: Why do we need the covector workflow for dev-refactor?

felsweg-iota: Agreed. That shouldn't be part of `dev-refactor`.

vuongDang (approved):

engine/new_runtime/src/memories/mod.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright 2020-2021 IOTA Stiftung
```
elenaf9: Nit: Move the mod file from `src/memories/mod.rs` -> `src/memories.rs` to be consistent in the module file naming throughout the crate.

engine/new_runtime/src/memories/buffer.rs
```diff
@@ -0,0 +1,375 @@
...
+    de::{Deserialize, Deserializer, SeqAccess, Visitor},
+    ser::{Serialize, SerializeSeq, Serializer},
+};
+
+/// GuardedMemory is used when we want to store sensitive non encrypted data
+/// This shall always be short lived
+#[derive(Clone, Eq)]
+pub struct Buffer<T: Bytes> {
+    boxed: Boxed<T>, // the boxed type of current GuardedVec
+}
```
elenaf9:
```suggestion
/// [`Buffer`] is used when we want to store sensitive non encrypted data.
/// This shall always be short lived.
#[derive(Clone, Eq)]
pub struct Buffer<T: Bytes> {
    boxed: Boxed<T>, // the boxed type of current Buffer
}
```

elenaf9: Also, maybe re-add the old comment about the security features and guarantees?: 
```suggestion
/// [`Buffer`] is used when we want to store sensitive non encrypted data.
/// This shall always be short lived.
///
/// Provides the following features and guarantees:
/// * Causes segfault upon access without using a borrow.
/// * Protected using mprotect:
///    * `Prot::NoAccess` - when the box has no current borrows.
///    * `Prot::ReadOnly` - when the box has at least one current immutable borrow.
///    * `Prot::ReadWrite` - when the box has a current mutable borrow (can only have one at a time).
/// * The allocated memory uses guard pages both proceeding and following the memory. Overflows and large underflows
///   cause immediate termination of the program.
/// * A canary proceeds the memory location to detect smaller underflows.  The program will drop the underlying memory
///   and terminate if detected.
/// * The Memory is locked with `mlock`.
/// * When the memory is freed, `munlock` is called.
/// * The memory is zeroed when no longer in use via `sodium_free`.
/// * `Buffer` types can be compared in constant time.
/// * `Buffer` types can not be printed using `Debug`.
/// `Buffer` includes serialization which converts the data into a vector before its serialized by serde.  Upon
/// deserialization, the data is returned back to a new `Buffer`.
#[derive(Clone, Eq)]
pub struct Buffer<T: Bytes> {
    boxed: Boxed<T>, // the boxed type of current Buffer
}
```

engine/new_runtime/src/memories/buffer.rs
```diff
@@ -0,0 +1,375 @@
...
+        Buffer {
+            boxed: Boxed::new(size, |b| b.as_mut_slice().copy_from_slice(payload)),
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.boxed.len()
+    }
+
+    #[allow(dead_code)]
```
elenaf9: Since `Buffer` is public I don't think we need  `#[allow(dead_code)]` for any of its methods.

engine/new_runtime/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
```
elenaf9: A `.license_template` is missing for `new_runtime`.

engine/new_runtime/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+serde = { version = "1.0", features = [ "derive" ] }
+random = { package = "rand", version = "0.8.4" }
+dirs = "4.0.0"
+
+[dependencies.iota-crypto]
+version = "0.8.0"
+features = [ "sha" ]
+
+[dev-dependencies]
+serde_json = "1.0"
```
elenaf9:
```suggestion
serde_json = "1.0"

```
Nit: newline at end of file

engine/new_runtime/src/memories/ram_memory.rs
```diff
@@ -0,0 +1,159 @@
...
+    de::{Deserialize, Deserializer, SeqAccess, Visitor},
+    ser::{Serialize, Serializer},
+};
+
+/// Protected ram memory that may be encrypted or not
+/// This is basically a wrapper for the Buffer type, but the usage
+/// is different, buffer type are meant for short lived usage while
+/// RamMemory can store data for longer period of time.
+/// Hence data in RamMemory has to be either encyrpted or protected
+/// behind a scheme
```
elenaf9:
```suggestion
/// Protected ram memory that may be encrypted or not.
/// This is basically a wrapper for the [`Buffer`] type, but the usage
/// is different. The `Buffer` type is meant for short lived usage while
/// `RamMemory` can store data for longer period of time.
/// Hence data in `RamMemory` has to be either encrypted or protected
/// behind a scheme.
```

engine/new_runtime/src/memories/ram_memory.rs
```diff
@@ -0,0 +1,159 @@
...
+            buf: Buffer::alloc(payload, size),
+            size,
+        })
+    }
+}
+
+impl LockedMemory for RamMemory {
+    /// Locks the memory and possibly reallocates
+    // Currently we reallocate a new RamMemory at each lock
+    // This improves security but decreases performance
```
elenaf9:
```suggestion
    /// Locks the memory and possibly reallocates.
    // Currently we reallocate a new [`RamMemory`] at each lock.
    // This improves security but decreases performance.
```

engine/new_runtime/src/memories/ram_memory.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+
+/// Protected ram memory that may be encrypted or not
+/// This is basically a wrapper for the Buffer type, but the usage
+/// is different, buffer type are meant for short lived usage while
+/// RamMemory can store data for longer period of time.
+/// Hence data in RamMemory has to be either encyrpted or protected
+/// behind a scheme
+#[derive(Clone)]
+pub struct RamMemory {
```
elenaf9:
```suggestion
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct RamMemory {
```

No need to manually implement `Debug`, `Serialize` and `Deserialize` below. They are already implemented on `Buffer`, so we can just derive them here.

engine/new_runtime/src/memories/ram_memory.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl Debug for RamMemory {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", DEBUG_MSG)
+    }
+}
+
+unsafe impl Send for RamMemory {}
+unsafe impl Sync for RamMemory {}
+
```
elenaf9:
```suggestion
```

`Send` and `Sync` are automatically derived since `Buffer` already implements them. Quoting the Rustonomicon:

> Send and Sync are also automatically derived traits. This means that, unlike every other trait, if a type is composed entirely of Send or Sync types, then it is Send or Sync. Almost all primitives are Send and Sync, and as a consequence pretty much all types you'll ever interact with are Send and Sync.

engine/new_runtime/src/memories/file_memory.rs
```diff
@@ -0,0 +1,333 @@
...
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+use zeroize::Zeroize;
+
+const FILENAME_SIZE: usize = 16;
+
+/// Data is stored into files in clear or encrypted.
+/// Basic security of this file includes files access control and
+pub struct FileMemory {
```
elenaf9: ...and what? :D

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+use MemoryShard::*;
+
+/// NonContiguousMemory only works on data which size corresponds to the hash primitive we use. In our case we use it to
+/// store keys hence the size of the data depends on the chosen box provider
+#[derive(Clone)]
+pub struct NonContiguousMemory {
+    shard1: MemoryShard,
+    shard2: MemoryShard,
+    config: NCConfig,
```
elenaf9: I don't think we need the `config` property here, it is only used on `alloc`. And if we need it again, we can just derive it by matching on the variants of `shard1` and `shard2`.

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,342 @@
...
+        match &self.shard1 {
+            RamShard(ram) => ram.unlock().expect("Failed to retrieve buffer from Ram shard"),
+            _ => unreachable!("{}", IMPOSSIBLE_CASE),
+        }
+    }
+
+    // Refresh the shards to increase security, may be called every _n_ seconds or
+    // punctually
+    #[allow(dead_code)]
+    fn refresh(self) -> Result<Self, MemoryError> {
```
elenaf9:
```suggestion
    pub fn refresh(self) -> Result<Self, MemoryError> {
```

Shouldn't this method be public to the user?

engine/new_runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -0,0 +1,342 @@
+// Copyright 2020-2021 IOTA Stiftung
```
elenaf9: Amazing work with the non-contiguous memory!

engine/src/vault/view.rs
```diff
@@ -353,13 +353,20 @@ impl Record {
...
+        let guarded = Buffer::alloc(&blob, tx.len.u64() as usize);
+
+        // let guarded =
+        //     GuardedVec::new(tx.len.u64() as usize, |i| {
+        //     let blob = SealedBlob::from(self.blob.as_ref())
+        //         .decrypt(key, tx.blob)
+        //         .expect("Unable to decrypt blob");
+
+        //     i.copy_from_slice(blob.as_ref());
+        // });
```
elenaf9, line 519:
```suggestion
```

Remove old code.

engine/src/vault/crypto_box.rs
```diff
@@ -51,37 +57,32 @@ impl<T: BoxProvider> Key<T> {
...
         } else {
             None
         }
     }
-
-    /// get the key's bytes from the [`GuardedVec`]
-    pub fn bytes(&self) -> Vec<u8> {
-        // hacks the guarded type.  Probably not the best solution.
-        (*self.key.borrow()).to_vec()
-    }
```
elenaf9, line 84: Why is this method removed? I think we need such a convenience method again, to avoid that the user has to the awkward call of:
```
let key = *(&key.key).borrow()
```
to access the actual key.

engine/src/vault/crypto_box.rs
```diff
@@ -36,11 +42,11 @@ pub trait BoxProvider: 'static + Sized + Ord + PartialOrd {
 }
 
 /// A key to the crypto box.  [`Key`] is stored on the heap which makes it easier to erase. Makes use of the
-/// [`GuardedVec<u8>`] type to protect the data.
+/// [`Buffer<u8>`] type to protect the data.
 #[derive(Serialize, Deserialize)]
 pub struct Key<T: BoxProvider> {
     /// the guarded raw bytes that make up the key
-    pub key: GuardedVec<u8>,
+    pub key: Buffer<u8>,
```
elenaf9, line 49: I know in practice it does not make a difference, but should we use the `RamMemory` type here because the `Key` is stored for a longer period of time? I thought the `Buffer` type should only be used when accessing a value in a `LockedMemory`.

engine/src/vault/crypto_box.rs
```diff
@@ -149,3 +151,154 @@ pub trait Decrypt<T: TryFrom<Vec<u8>>>: AsRef<[u8]> {
...
+    pub key: NonContiguousMemory,
+
+    /// phantom data to call to the provider.
+    #[serde(skip_serializing, skip_deserializing)]
+    _box_provider: PhantomData<T>,
+}
+
+impl<T: BoxProvider> NCKey<T> {
+    /// generate a random key using secure random bytes
+    #[allow(dead_code)]
```
elenaf9: Shouldn't `engine/lib.rs` export this type? In which case we also don't need `#[allow(dead_code)]` on any of its public mehtods.

engine/src/vault/key_store.rs
```diff
@@ -0,0 +1,91 @@
...
+use std::collections::HashMap;
+
+/// The [`KeyStore`] keeps a map of [`VaultId`] -> [Vec<u8>] representing
+/// encrypted [`Key<P>`] using the `master_key`.
+/// `master_key` is stored in a non-contiguous data structure [`NCKey<P>`]
+/// for more security
+pub struct KeyStore<P: BoxProvider> {
+    store: HashMap<VaultId, Vec<u8>>,
+    master_key: NCKey<P>,
+}
```
elenaf9: Nice, this also solves issue #268!

engine/src/vault/key_store.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl<P: BoxProvider> KeyStore<P> {
+    #![allow(dead_code)]
```
elenaf9: Again, I think `KeyStore` should also be public outside of the crate so we can use it in `iota_stronghold`. In that case we also don't need the `#![allow(dead_code)]`.

elenaf9 (approved):
Really great work!

Sorry for my late review here, did not have time to look through #316 before it was merged.
Couple of comments. Almost all are nitpicks, the only blocking ones are that `engine::vault::{KeyStore, NCKey}` should be public if we want to use them in the client.

engine/src/vault/key_store.rs
elenaf9: Resolved with #324.

engine/src/vault/crypto_box.rs
elenaf9: Resolved with #324.

engine/new_runtime/src/memories/file_memory.rs
tensor-programming: haha

client_new/src/types/stronghold/base.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{Client, ClientError, ClientState, KeyProvider, LoadFromPath, Snapshot, SnapshotPath, Stronghold, UseKey};
+use engine::vault::ClientId;
+use std::ops::Deref;
+
+impl Stronghold {
```
tensor-programming: I know these aren't the complete interfaces for the user to interact with, but maybe it would be a little bit more clear how a user should interact with this library if the stronghold type and interfaces weren't nested so far in a folder like this.  The same for the other front facing types, client, keystore etc.  Or at least some docs that highlight that they are the main types that should be used.

client_new/src/tests/interface_tests.rs
```diff
@@ -0,0 +1,118 @@
+// Copyright 2020-2021 IOTA Stiftung
```
tensor-programming: I wonder, do these tests even need to be a part of the core program anymore or can they be put into a test file?  The reason before was because there were some items things that weren't public which needed to be tested through the interface.

felsweg-iota: could be moved into an integration tests, yes

client_new/src/types/stronghold/base.rs
felsweg-iota: yeah, docs coming up. i mean for the `Stronghold` type the intention was to separate `base` impl. for all local  and `p2p` for all (upcoming / ported) networking functionality.

engine/new_runtime/src/memories/file_memory.rs
felsweg-iota: you're so mean :rofl:

client_new/src/types/client.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// ```
+    /// ```
+    pub fn vault(&self, vault_path: Location) -> ClientVault {
+        let (vault_id, _) = vault_path.resolve();
+
+        ClientVault {
+            client: self.atomic_ref(),
+            id: vault_id,
+        }
+    }
```
elenaf9: With #319 all operations on the vault can be done through procedures. We could consider removing `ClientVault` so that we have a single way of doing things through procedures only.

client_new/src/types/client.rs
```diff
@@ -0,0 +1,207 @@
...
+        let mapped = res.map(|mut vec| vec.pop().unwrap().try_into().ok().unwrap())?;
+        Ok(mapped)
+    }
+
+    /// Executes a list of cryptographic [`Procedures`] sequentially and returns a collected output
+    ///
+    /// # Example
+    /// ```no_run
+    /// ```
+    pub async fn execure_procedure_chained(
```
elenaf9: Why are some of the methods on the `Client` async? Unless I am missing something, there is no need for it.

client_new/src/types/client.rs
felsweg-iota: `ClientVault` is a thin wrapper around `Client` and calls all actions directly on `Client`s database.

elenaf9: Yes but said actions (`write_secret`, `delete_secret`, `revoke_secret` and `cleanup`) can all be done through procedures (`StrongholdProcedure::WriteVault`, `StrongholdProcedure::RevokeData`, etc). 
Thus the methods are kinda redundant and if we would remove them, we also would not need this `ClientVault` wrapper anymore.

felsweg-iota: I disagree. The `ClientVault` is meant to have as a distinct type to differentiate the actions taken on them. The whole idea of the new client interface is around types.

felsweg-iota: The only point to argue would be to expose this `ClientVault` as an interface on `Client`.

client_new/src/types/client.rs
felsweg-iota: true, it's a reminiscent of the old interface. Most of the operations are sync, since we do not enforce a hard dependency on some async runtime. I/O operations on the `Snapshot` should be async, imho. I wonder, if we should enforce a dependency on tokio (or any other async runtime)

client_new/src/types/client.rs
elenaf9: My argument was based on what is described in the _Primitive Based Interface_ RFC in #313 in **Reference-level explanation - Vault**: 

> While the vault is a stronghold primitive, it does not have its own type (being the only exception). Most interesting operations on a vault are procedures. Since a client can have multiple vaults, a procedure generally works on the client-level rather than the vault-level (...)

I think it's cleaner to have a unified way of interacting with vaults by always using procedures. It's also easier for the user to understand, rather than having multiple ways of how something can be written to a vault.

But I also see your point. Having a `ClientVault` type is more consistent with how the interface is generally designed.

Anyway, your call 😄.

felsweg-iota: The execution of procedures are on `Client` level, while `ClientVault` offers a more distinct type :D Sure, every action can be done by procedures, but having access to "simpler" function calls might be beneficial, imho.

engine/src/vault/crypto_box.rs
vuongDang: Yes you're right, will change it

PR #321, opened 2022-02-25
fix(*): fix clippy v0.1.59

elenaf9:
# Description of change

Fix lints from new Clippy version 0.1.59:
- clippy::needless-late-init
- clippy::unnecessary-to-owned
- clippy::map-flatten


felsweg-iota (approved):
good idea to have it in another PR.

PR #325, opened 2022-03-02
GitHub workflow: add rustdoc check

elenaf9:
# Description of change

Add GitHub Workflow for checking Rust intra-doc links.

Fix broken intra-doc links.

PR #326, opened 2022-03-03
GitHub workflow: add engine/runtime to clippy & fmt check

elenaf9:
# Description of change

While working on #325 I noticed that the runtime is not included in the clippy and format workflows, because it is its own crate and not part of the engine.
This PR adds it to the clippy and format checks.

@nothingismagick  was there a specific reason that is was originally not included / do you see an issue with this change?


felsweg-iota (approved):
LGTM.

PR #327, opened 2022-03-06
Client/P2p: refactor firewall usage, add client mapping

elenaf9:
# Description of change

Refactor the integration of the p2p firewall in the client:
- Implement the `Rule::Restricted` in the Stronghold client's firewall: 
  - `AccessRequest` derives for a `ShRequest` the vaults that are targeted and the type of access(es) that it needs:
    a) Use a secret from the vault (e.g. to sign a message)
    b) Write to the vault
    c) Clone secrets from the vault (-> Sync)
- `Permissions` can be set by a user to restrict what type of access to what vaults and clients are allowed
  -> for each inbound `ShRequest` the `AccessRequest` is derived, and then approved or rejected according to the set `Permissions`
- Add client-path to `ShRequest`
  - Forward the request at the remote to this client, instead of the current "target" client 
  - <s>Allow a the remote to configure a mapping from the "requested" client path to a local client</s>
  _After rethinking this, I don't think there is a real use case for such a mapping. Instead, it just add unnecessary complexity and is especially difficult to integrate this feature within the firewall. Thus I removed this feature again in 6ba31ac._
- Separate the p2p logic in the client into `state::p2p` and `actor::p2p` to separate the actix-related logic from -independent one 
  -> actix-independent code can be kept after we remove actix  

In the `stronghold_p2p `crate: 
- Remove generic parameter `Fn(&TRq) -> bool` from `p2p::firewall::Rule` , instead wrap it in a `Arc<dyn ..>` directly in the `Rule::Restricted` variant 
- Add `FwRequest` trait as trait bound to generic `TRq` instead of `Borrow<_>`
- Upgrade libp2p from v0.41.0 to v0.43.0
- Rename `SwarmTask` -> `EventLoop`
- Clean code of `behaviour::request_manger`
- Removed firewall checks on outbound requests

#### TODO:
- [x] Add tests for `Stronghold` interface
   - [x] Firewall
   - [x] <s> Client Mapping </s>
- [x] Optionally expose the firewall channel in `Stronghold` interface for asynchronous firewall interaction
- [x] Add doc examples
- [x] <s>  Consider potential client mapping when applying firewall rules </s>

## Type of change

Adjusted existing tests in `stronghold_p2p`.
Moved `iota_stronghold` p2p tests in Stronghold interface to new module, added `test_p2p_firewall`.

## Change checklist


client/src/procedures/primitives.rs
```diff
@@ -77,6 +77,27 @@ impl Procedure for StrongholdProcedure {
 }
 
 impl StrongholdProcedure {
+    pub(crate) fn input(&self) -> Option<Location> {
+        match self {
+            StrongholdProcedure::CopyRecord(CopyRecord { source: input, .. })
```
felsweg-iota, line 82: I encountered similar matching pattern for enums, so I wonder if this can be done in a more elegant way.

felsweg-iota (approved):
Excellent work! I think the update makes the integration of the firewall and related components much cleaner.

client/src/procedures/primitives.rs
elenaf9: Agree, that would be desirable. But I was not able to come up with a better solution. Maybe there is a crate that provides a macro for something like this?

elenaf9: Note: I merged the PR now without changing this. If we come up with a solution for this in the future we can do it in a separate PR and apply it on all locations where it is needed, e.g. also in the case that you encountered.

PR #330, opened 2022-03-20
feat(engine): low-level logic for merging vaults

elenaf9:
# Description of change

Extract from #317 the changes that concern the engine. 
The added methods will enable the synchronization of vaults in the `client`:
- list existing entries with id and blob-id
- changing encryption key and/ or id of a record
- exporting and importing records from / to a vault
- the blob-id now changes each time the data is updated

The actual logic for state-synchronization of clients and strongholds will follow in #317 or a new separate PR.

## Type of change

## Change checklist

### DIFF
--- a/client/src/actors/secure.rs
+++ b/client/src/actors/secure.rs
@@ -305,15 +305,7 @@ impl_handler!(messages::ClearCache, (), (self, _msg, _ctx), {
 
 impl_handler!(messages::CheckRecord, bool, (self, msg, _ctx), {
     let (vault_id, record_id) = msg.location.resolve();
-
-    return match self.keystore.take_key(vault_id) {
-        Some(key) => {
-            let res = self.db.contains_record(&key, vault_id, record_id);
-            self.keystore.insert_key(vault_id, key);
-            res
-        }
-        None => false,
-    };
+    self.db.contains_record(vault_id, record_id)
 });
 
 impl_handler!(messages::WriteToVault, Result<(), RecordError>, (self, msg, _ctx), {
--- a/engine/src/vault.rs
+++ b/engine/src/vault.rs
@@ -24,6 +24,6 @@ pub mod view;
 pub use crate::vault::{
     base64::{Base64Decodable, Base64Encodable},
     crypto_box::{BoxProvider, Decrypt, DecryptError, Encrypt, Key, NCDecrypt, NCEncrypt, NCKey},
-    types::utils::{ChainId, ClientId, Id, InvalidLength, RecordHint, RecordId, VaultId},
+    types::utils::{BlobId, ChainId, ClientId, Id, InvalidLength, RecordHint, RecordId, VaultId},
     view::{DbView, RecordError, VaultError},
 };
--- a/engine/src/vault/crypto_box.rs
+++ b/engine/src/vault/crypto_box.rs
@@ -129,8 +129,8 @@ impl<T: BoxProvider> Debug for Key<T> {
 /// trait for encryptable data. Allows the data to be encrypted.
 pub trait Encrypt<T: From<Vec<u8>>>: AsRef<[u8]> {
     /// encrypts a raw data and creates a type T from the ciphertext
-    fn encrypt<B: BoxProvider, AD: AsRef<[u8]>>(&self, key: &Key<B>, ad: AD) -> Result<T, B::Error> {
-        let sealed = B::box_seal(key, ad.as_ref(), self.as_ref())?;
+    fn encrypt<P: BoxProvider, AD: AsRef<[u8]>>(&self, key: &Key<P>, ad: AD) -> Result<T, P::Error> {
+        let sealed = P::box_seal(key, ad.as_ref(), self.as_ref())?;
         Ok(T::from(sealed))
     }
 }
--- a/engine/src/vault/view.rs
+++ b/engine/src/vault/view.rs
@@ -109,13 +109,26 @@ impl<P: BoxProvider> DbView<P> {
         }
     }
 
+    /// Check to see if a vault with the given [`VaultId`] is present.
+    pub fn contains_vault(&self, vid: &VaultId) -> bool {
+        self.vaults.contains_key(vid)
+    }
+
     /// Check to see if a [`Vault`] contains a [`Record`] through the given [`RecordId`].
-    pub fn contains_record(&mut self, key: &Key<P>, vid: VaultId, rid: RecordId) -> bool {
+    pub fn contains_record(&self, vid: VaultId, rid: RecordId) -> bool {
         if let Some(vault) = self.vaults.get(&vid) {
-            vault.contains_record(key, rid)
+            vault.contains_record(rid)
         } else {
             false
         }
+    }
+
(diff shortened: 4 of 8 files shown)


client/src/actors/secure.rs
```diff
@@ -305,15 +305,7 @@ impl_handler!(messages::ClearCache, (), (self, _msg, _ctx), {
...
-
-    return match self.keystore.take_key(vault_id) {
-        Some(key) => {
-            let res = self.db.contains_record(&key, vault_id, record_id);
-            self.keystore.insert_key(vault_id, key);
-            res
-        }
-        None => false,
-    };
+    self.db.contains_record(vault_id, record_id)
```
tensor-programming, line 308: As an aside, love that you guys got rid of the `take_key`/`insert_key` stuff.

engine/src/vault/view.rs
```diff
@@ -183,6 +196,82 @@ impl<P: BoxProvider> DbView<P> {
...
+
+    /// List the ids of all records in the vault.
+    pub fn list_records(&self, vid: &VaultId) -> Vec<RecordId> {
+        self.vaults
+            .get(vid)
+            .map(|v| v.entries.keys().map(|&c| c.into()).collect())
+            .unwrap_or_default()
+    }
+
+    /// List [`RecordId`] and [`BlobId`] of all entries in the vault.
```
tensor-programming, line 213: I wonder if its ok to expose the `BlobId`.  Before this was just an internal reference to connect the transaction head to the record; is this even useful for external resources?

engine/src/vault/view.rs
```diff
@@ -183,6 +196,82 @@ impl<P: BoxProvider> DbView<P> {
...
+                let entries = vault.entries.iter().map(|(&id, r)| (id.into(), r.clone())).collect();
+                (*vid, entries)
+            })
+            .collect()
+    }
+
+    /// Clone the specified records from the vault without removing them.
+    ///
+    /// **Note:** before importing these records to a new vault with [`DbView::import_records`], [`Record::update_meta`]
+    /// has to be called on each [`Record`] to re-encrypt it with the target vault's encryption key.
```
tensor-programming, line 239: What happens if you export/import a record with the 'wrong' encryption key?  Is it just unreadable or is there an error?

engine/src/vault/view.rs
```diff
@@ -400,6 +537,46 @@ impl Record {
         Ok(())
     }
 
+    /// Update the key and id of an existing [`Record`].
```
tensor-programming, line 540: This looks fine but I do wonder if there is some way that we could introduce some obfuscation/entropy in here.  In essence, this is a read function for the record; where you decrypt a record (read it) and then re-encrypt it.   Obviously, if we are going to rekey the data, there really isn't a easy way around revealing the record even if its only for a couple of microseconds.  I almost wonder if we should explore something like proxy re-encryption to avoid exposing the secret at all.

tensor-programming (approved):

engine/src/vault/view.rs
elenaf9: The `BlobId` is exposed so that we can check if the content of a record has changed, without needing to expose the actual secret. Each time we update its content, we create a new random `BlobId` for this record. This is useful for when we sync two vaults multiple times. To reduce overhead and risk, we only want to import the records that we don't have yet, or whose content has changed. For this we compare the `BlobId` on source and target, and only import the record if it is different.

engine/src/vault/view.rs
elenaf9: It will cause a `RecordError` once you then try to read/ use the record, because the vault-key won't match. I originally implemented it so that the user has to provide the old encryption key and that the records are then re-encrypted with the new key on import. The problem is that on the higher layer on sync in #317with the mapper  it currently allows to re-arrange records to different locations in a very flexible way. With that, it could be that the list of records come from multiple different vaults and/ or clients. In that case we don't have a single "old" encryption key, but a map of keys, depending what the "old" location of the record was. Therefore I decided to move that logic into the sync-implementation. Will think about this a bit more and maybe change it again once I refactor the client-layer part of the sync feature.

engine/src/vault/view.rs
elenaf9: Definitely sounds worth exploring. I would still already merge this into `dev-refactor` so that it can be included in the security audit starting today, But I will follow up on that idea and look into proxy re-encryption and how we can use it here, before it is merged to `dev`.

engine/src/vault.rs
```diff
@@ -24,6 +24,6 @@ pub mod view;
 pub use crate::vault::{
     base64::{Base64Decodable, Base64Encodable},
     crypto_box::{BoxProvider, Decrypt, DecryptError, Encrypt, Key, NCDecrypt, NCEncrypt, NCKey},
-    types::utils::{ChainId, ClientId, Id, InvalidLength, RecordHint, RecordId, VaultId},
```
felsweg-iota, line 27: I wonder, if `ChainId` can be renamed to `RecordId` as it is used as such.

engine/src/vault/crypto_box.rs
```diff
@@ -129,8 +129,8 @@ impl<T: BoxProvider> Debug for Key<T> {
 /// trait for encryptable data. Allows the data to be encrypted.
 pub trait Encrypt<T: From<Vec<u8>>>: AsRef<[u8]> {
     /// encrypts a raw data and creates a type T from the ciphertext
-    fn encrypt<B: BoxProvider, AD: AsRef<[u8]>>(&self, key: &Key<B>, ad: AD) -> Result<T, B::Error> {
-        let sealed = B::box_seal(key, ad.as_ref(), self.as_ref())?;
+    fn encrypt<P: BoxProvider, AD: AsRef<[u8]>>(&self, key: &Key<P>, ad: AD) -> Result<T, P::Error> {
```
felsweg-iota, line 132: P > B

felsweg-iota (approved):
Looks good to me. Let's also see, if we can introduce some proxy- encryption for updating meta data.

engine/src/vault.rs
elenaf9: I was wondering about that as well. I think renaming `ChainId` to `RecordId` would break the old snapshot format, but there is probably a way to solve that.

engine/src/vault/view.rs
vuongDang: I was confused about the BlockId, too.
I think we should definitely think about renaming certain fields for clarity without breaking breaking the snapshot format

vuongDang (commented):
Looks good to me

engine/src/vault.rs
tensor-programming: Unless something else has changed, the `ChainId` and the `RecordId` should be the same thing; even just rewrapping the `ChainId` to be a `RecordId` would be possible.  The concept of the `ChainId` doesn't make too much sense anymore since the Records aren't Chains of versions anymore.

engine/src/vault/crypto_box.rs
tensor-programming: I do agree, though I did use B for Box, it makes a bit more sense to use P for Provider.

PR #334, opened 2022-03-29
Feat(client_new): State synchronization of Clients and Snapshots

### LINKED ISSUE #278: State Synchronization Across Multiple Stronghold Instances

elenaf9:
# Description of change

Add the synchronization feature to the new client.
This is a refactor of #317, ported to the new client. 

Further improvements included in this PR outside of the sync feature:
- remove unnecessary `mut` requirement from method signatures
- favor `KeyStore::get_key` over `KeyStore::take_key` where possible

Remaining TODOs for this PR:
- [x] handle errors
- [x] add docs

Remaining TODOs outside of this PR:
- Add remote sync
- Test (remote) snapshot sync

// CC @PhilippGackstatter: Can you confirm that your use-case in identity is solved with [`Client::sync_with`](https://github.com/iotaledger/stronghold.rs/blob/d427893d73ed3a96dff1d271bada1edba7fc8a4b/client_new/src/types/client.rs#L159-L164)?

Fixes issue #278.

## Type of change

Added test for client-sync.

## Change checklist

### DIFF
--- a/client_new/src/security/keystore.rs
+++ b/client_new/src/security/keystore.rs
@@ -44,6 +44,12 @@ where
         let enc_key = self.store.remove(&id)?;
         self.master_key.decrypt_key(enc_key, id).ok()
     }
+    /// Gets the encrypted key from the [`KeyStore`].
+    /// Decrypt it with the `master_key` and `vault_id` as salt.
+    pub fn get_key(&self, id: VaultId) -> Option<Key<P>> {
+        let enc_key = self.store.get(&id)?.clone();
+        self.master_key.decrypt_key(enc_key, id).ok()
+    }
 
     /// Checks to see if the vault exists.
     pub fn vault_exists(&self, id: VaultId) -> bool {
@@ -61,7 +67,7 @@ where
     /// Inserts a key into the [`KeyStore`] by [`VaultId`].
     /// If the [`VaultId`] already exists, it just returns the existing [`Key<P>`]
     pub fn get_or_insert_key(&mut self, id: VaultId, key: Key<P>) -> Result<Key<P>, P::Error> {
-        let vault_key = if let Some(key) = self.take_key(id) { key } else { key };
+        let vault_key = if let Some(key) = self.get_key(id) { key } else { key };
         let enc_key = self.master_key.encrypt_key(&vault_key, id)?;
         self.store.insert(id, enc_key);
         Ok(vault_key)
--- a/client_new/src/security/keystore.rs
+++ b/client_new/src/security/keystore.rs
@@ -87,7 +87,7 @@ where
     pub fn rebuild_keystore(&mut self, keys: HashMap<VaultId, Key<P>>) -> Result<(), P::Error> {
         let mut new_ks = KeyStore::default();
         for (id, key) in keys.into_iter() {
-            new_ks.get_or_insert_key(id, key)?;
+            new_ks.insert_key(id, key)?;
         }
         *self = new_ks;
         Ok(())
--- a/client_new/src/sync.rs
+++ b/client_new/src/sync.rs
@@ -19,10 +19,17 @@ pub enum MergePolicy {
     Replace,
 }
 
+impl Default for MergePolicy {
+    fn default() -> Self {
+        MergePolicy::Replace
+    }
+}
+
 // TODO: Use *_paths instead of `Ids`.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct SyncClientsConfig {
     /// Optionally only perform a partial sync with specific vaults.
+    /// In case of `None` all vaults are synched.
     ///
     /// Note: This is referring to the Ids as they are on the source client, not
     /// to the mapped id.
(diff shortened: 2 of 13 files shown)


client_new/src/security/keystore.rs
```diff
@@ -81,7 +87,7 @@ where
     pub fn rebuild_keystore(&mut self, keys: HashMap<VaultId, Key<P>>) -> Result<(), P::Error> {
         let mut new_ks = KeyStore::default();
         for (id, key) in keys.into_iter() {
-            new_ks.get_or_insert_key(id, key)?;
+            new_ks.insert_key(id, key)?;
```
felsweg-iota, line 90: :+1:

client_new/src/types/client.rs
```diff
@@ -107,19 +109,95 @@ impl Client {
...
+    /// If `select_records` is `Some` only the specified records are copied, else a full sync
+    /// is performed. If a record already exists at the target, the [`MergePolicy`] applies.
+    ///
+    /// # Example
+    /// ```
+    /// ```
+    pub fn sync_vaults(
+        &self,
+        source: VaultId,
+        target: VaultId,
```
PhilippGackstatter: It looks like this is a public function, so shouldn't this take `Vec<u8>` vault paths instead of the (basically internal) `VaultId` type?

client_new/src/types/location.rs
```diff
@@ -97,7 +97,17 @@ where
 }
 
 // Derives the counter [`RecordId`] from the given vault path and the counter value.
-pub fn derive_record_id<P>(vault_path: P, counter: usize) -> RecordId
+pub fn derive_record_id<V, R>(vault_path: V, record_path: R) -> RecordId
+where
+    V: AsRef<Vec<u8>>,
+    R: AsRef<Vec<u8>>,
+{
```
PhilippGackstatter: I'm seeing this `AsRef<Vec<u8>>` pattern frequently in stronghold, but I don't understand it. Why not take `AsRef<[u8]>` instead, which would work just as well? The `Vec<u8>` approach forces users to construct a vector, even though a slice of bytes would technically be sufficient. `Client::vault_exists` is one such example where passing a `&[u8]` wouldn't require any internal changes and be helpful to the caller, so one doesn't have to allocate a vector.

PhilippGackstatter (commented):
`Client::sync_with` works perfectly for us, thanks for the ping!

Just commented on some things I noticed while browsing the code.

felsweg-iota:
@PhilippGackstatter 
> I'm seeing this AsRef<Vec<u8>> pattern frequently in stronghold, but I don't understand it. Why not take AsRef<[u8]> instead, which would work just as well? The Vec<u8> approach forces users to construct a vector, even though a slice of bytes would technically be sufficient. Client::vault_exists is one such example where passing a &[u8] wouldn't require any internal changes and be helpful to the caller, so one doesn't have to allocate a vector.

The reasoning behind this (minimal)  decision is, that `Location` stores a vector. But I do see the point, that `AsRef<[u8]>` is easier to construct from a user perspective. I think that's easy to adapt.

PhilippGackstatter:
> The reasoning behind this (minimal) decision is, that Location stores a vector. But I do see the point, that AsRef<[u8]> is easier to construct from a user perspective. I think that's easy to adapt.

For `Location` that's fine. The `Location` constructors already take `Into<Vec<u8>>` which makes sense. But for deriving vault, record or client ids, only a slice is necessary internally, so it doesn't make sense to force users to supply a vector.

felsweg-iota (commented):
Looks good to me. I will take over the changes proposed by @PhilippGackstatter

client_new/src/types/client.rs
elenaf9: Good catch, yes. Fixed with 35c7402.

client_new/src/types/location.rs
elenaf9: Good idea! Changed trait bound to `AsRef<[u8]>` 1c28135.

client_new/src/types/client.rs
PhilippGackstatter: Perhaps `VaultId` and other id types should be made private, so the compiler would flag things like these, just a thought.

elenaf9:
Another note because I re-arranged the logic for resolving \*paths to \*Ids (in particular vault-paths, records-paths and `Location`s):
 @felsweg-iota  could you please double-check again that I did not break how paths are resolved? I checked it myself but it would be good if we put special attention on this. It would break backward-compatibility if I accidentally did.

client_new/src/sync.rs
```diff
@@ -0,0 +1,704 @@
...
+        let v_path = random::bytestring(4096);
+        let r_path = random::bytestring(4096);
+        Location::generic(v_path, r_path)
+    }
+
+    fn vault_path_to_id(path: &str) -> VaultId {
+        derive_vault_id(path.as_bytes())
+    }
+
+    fn r_ctr_to_id(vault_path: &str, ctr: usize) -> RecordId {
```
felsweg-iota, line 425: r2d2

felsweg-iota (approved):
looks good to me

PR #337, opened 2022-04-12
Network Client Port

felsweg-iota:
# Description of change

This PR ports the former network client to the new client interface. Operating with remote Strongholds is done via distinct types. The main binding type is [`Peer`]. 

Note: The current PR is a draft and has not all functionality ported to the new client interface.

None.

## Type of change

The code has been checked for validity through integration tests. 

## Change checklist


client_new/src/types/store.rs
```diff
@@ -87,13 +71,13 @@ impl Store {
...
-    pub fn get(&self, key: Vec<u8>) -> Result<StoreGuard<'_>, ClientError> {
-        let guard = self.cache.try_read()?;
+    pub fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, ClientError> {
+        let guard = self.cache.try_read().map_err(|_| ClientError::LockAcquireFailed)?;
 
         // Problem: The returned rwread guard is local to this function, hence we can't return a borrowed ref
         // to the inner value. we could return the guard itself, but would rely on the user to deref the rwguard
         // and then access the value again
-        Ok(StoreGuard::from(guard, key))
+        Ok(guard.get(&key.to_vec()).cloned())
```
PhilippGackstatter, line 80: Thanks for changing this to take `&[u8]`. You can also get rid of the internal clone by changing the definition of `Cache::get` to match that of `HashMap::get`:

```rust
impl Cache {
  pub fn get<Q>(&self, key: &Q) -> Option<&V>
      where
          K: std::borrow::Borrow<Q>,
          Q: ?Sized + Hash + Eq,
      { ... }
}
```

Then you can simply do:

```suggestion
        Ok(guard.get(key).cloned())
```

The same probably applies to `remove` and perhaps other `Cache` operations that just wrap `HashMap`.

client_new/src/types/store.rs
```diff
@@ -87,13 +71,13 @@ impl Store {
     /// assert!(store.get(key.clone()).is_ok());
     /// assert!(store.get(key).unwrap().deref().is_some());
     /// ```
-    pub fn get(&self, key: Vec<u8>) -> Result<StoreGuard<'_>, ClientError> {
-        let guard = self.cache.try_read()?;
+    pub fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, ClientError> {
```
PhilippGackstatter, line 74: I thought it was an excellent feature that this returned a reference to some bytes and avoided a clone. In our particular case, we actually never need a copy of the data, because we deserialize the return value into something else, and a reference is sufficient for that.

client_new/src/types/stronghold.rs
```diff
@@ -28,11 +63,626 @@ use crate::{Client, ClientError, Snapshot};
...
+
+        Ok(client)
+    }
+
+    /// Writes all client states into the [`Snapshot`] file
+    ///
+    /// # Example
+    /// ```
+    /// ```
+    pub fn commit(&self, snapshot_path: &SnapshotPath, keyprovider: &KeyProvider) -> Result<(), ClientError> {
```
PhilippGackstatter, line 264: If this is not `async`, then this will block the current thread while waiting for I/O to complete? In that case, all async tasks running on this thread would be blocked. I think you mentioned before that the locks in the `Client` are temporary until the RLU is hopefully working. Is that also the case here? Is this how it's supposed to work, or is that also temporary? Because writing the snapshot should definitely use async I/O in my opinion.
I would argue the same for `Stronghold::load_snapshot`.

client_new/src/tests/network_tests.rs
```diff
@@ -0,0 +1,815 @@
...
+        assert!(result.is_ok(), "Assertion Failed= {:?}", result);
+
+        // create peer
+        let result = local.create_remote_client(remote_public_key, remote_client_path).await;
+        assert!(result.is_ok(), "Assertion Failed= {:?}", result);
+
+        // connect peer
+        let peer = result.unwrap();
+        let result = peer.connect().await;
+        assert!(result.is_ok(), "Assertion Failed= {:?}", result);
```
PhilippGackstatter, line 614: Why can't `connect` be a private method and be called from within `create_remote_client`? This seems like an unnecessary extra step. What if I call `peer.remote_write_store` before calling `peer.connect`? I assume I get an error? 
Would be nice if obtaining an instance of a `Peer` would mean that I'm in a connected state right from the start.

PhilippGackstatter (commented):
I don't think I have the capacity to dive into the network part fully, but it looks good to me from a high-level. I think I can give better feedback when I actually use an interface to implement something.

Thanks for addressing the feedback for the non-p2p parts. I still have some wishes regarding that :slightly_smiling_face: 

- `SnapshotPath` should implement `Display`.
  - It can delegate to the internal `Path` display impl. We use this in an error and it's nice to be able to print the path in non-Debug format.
- `Stronghold` and `KeyProvider` should implement `Debug` (see also the interoperability and debuggability section in the [Rust API guidelines](https://rust-lang.github.io/api-guidelines/checklist.html)). It can be a conservative implementation that doesn't print any secrets of course.
- `RecordHint` should be removed from all `procedures::*` types, so users don't have to depend on the engine crate for just that type (because it's (correctly) no longer re-exported from the interface).
- `stronghold_engine::new_runtime::MemoryError` should be re-exported from the interface so users don't have to depend on the engine crate, and that type should implement `std::error::Error`.

client_new/src/types/store.rs
felsweg-iota: I agree, but there were some problems with the lifetime being used in an async context. But I can look into that again. 

The intend to return a cloned copy was, that returning copies of the Store wouldn't cost much and do not contain any sensitive data.

client_new/src/types/stronghold.rs
felsweg-iota: I agree. IO  should be async. When this PR is feature complete, I will tackle the remaining issues.

client_new/src/tests/network_tests.rs
felsweg-iota: The idea was to have two separate atomic operations, one to obtain a saved peer, the other to connect. Only a `disconnect` function is missing. I can evaluate if a `Peer` can be connected right away

felsweg-iota:
@PhilippGackstatter Thanks for the comments! Most of what you mentioned is on the list of changes.

client_new/src/types/store.rs
PhilippGackstatter: It's definitely nice to have, but I suppose getting the interface architecture to a more stable point is more important.

PhilippGackstatter:
Two more things I found:

- Looks like `Client::record_exists` could take `&Location` instead of `Location`.
- This is really not a priority, but perhaps it would also be feasible to make the `procedures::*` generic over a lifetime. That way they wouldn't have to take owned parameters like `Location` in `procedures::GenerateKey`, further reducing the number of clones.

felsweg-iota:
@PhilippGackstatter concerning 
> stronghold_engine::new_runtime::MemoryError should be re-exported from the interface so users don't have to depend on the engine crate, and that type should implement std::error::Error.

Even though parts of the new runtime make use of `std` features, the main goal is to explicitly opt-in for `std` as already done in the client. I do see the point to implement the standard Error.

client_new/src/types/snapshot.rs
```diff
@@ -129,6 +119,12 @@ impl SnapshotPath {
     }
 }
 
+impl Display for SnapshotPath {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "SnapshotPath: {:?}", self.path)
+    }
+}
+
```
PhilippGackstatter: Nit: This is currently basically identical to the `Debug` impl. When printing an error message that includes a `SnapshotPath`, the reader probably doesn't care that this is a `SnapshotPath`, the actual inner path would be sufficient. For `Debug` context is good, for `Display` it's not necessary.

```suggestion
impl Display for SnapshotPath {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.path.display())
    }
}

```

PR #347, opened 2022-04-27
Add the `ConcatKdf` procedure

### LINKED ISSUE #344: [Request] Add a procedure that executes the concat KDF
## Description

For DIDComm communication, the identity team requires the use of `ECDH-ES` in combination with some key wrapping algorithm like `A256KW`. The latter was requested in #338, while the former requires #339 as well as the concat KDF, defined in [NIST SP 800-56A section 5.8.1](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-56ar.pdf#page=47). As an example context, this will be used according to [RFC 7518: JSON Web Algorithms](https://www.rfc-editor.org/rfc/rfc7518.html#section-4.6.2).

An implementation, that should be copyable, can be found here: https://github.com/iotaledger/identity.rs/blob/4287c9834de8dc42fd4ff0c0e8b3a2bc822d0ec4/libjose/src/utils/crypto/concat_kdf.rs.

This needs to be a stronghold procedure, because this KDF works on secrets which were previously created through diffie hellman or concatenation (#339).

## Motivation

Implementing DIDComm encryption in identity.rs.

## Requirements

1. Expose a procedure that implements the concat KDF.

I'm open to it, just let me know.

PhilippGackstatter:
# Description of change

Adds the concat KDF as a procedure. See #344 for more details.

Also pulls in some parts from the old stronghold tests to implement a test for the newly introduced functionality. Notably, this also implements `ClientVault::read_secret` for test purposes and is thus behind the `#[cfg(test]` feature.

fixes #344 

## Type of change

- [ ] Documentation Fix

Added a test similar to the existing X25519 + HKDF test but using the concat KDF instead.

A [test vector from RFC 7518: JSON Web Algorithms](https://www.rfc-editor.org/rfc/rfc7518.html#appendix-C) was added in a unit test.

## Change checklist


client/src/procedures/primitives.rs
```diff
@@ -789,3 +798,115 @@ impl UseSecret for AeadDecrypt {
...
+            digest.update(&(apu.len() as u32).to_be_bytes());
+            digest.update(apu);
+
+            // PartyVInfo
+            digest.update(&(apv.len() as u32).to_be_bytes());
+            digest.update(apv);
+
+            // Shared Key Length
+            digest.update(&((len * 8) as u32).to_be_bytes());
+
```
cycraig: I can't find a reference to including the output length as part of the hash in the NIST PDF (it's not part of `OtherInfo` as far as I can see), nor any existing implementations such as Hyperledger Aries: https://github.com/hyperledger/aries-askar/blob/main/askar-crypto/src/kdf/concat.rs

client/src/procedures/primitives.rs
```diff
@@ -789,3 +798,115 @@ impl UseSecret for AeadDecrypt {
...
+    }
+
+    fn target(&self) -> &Location {
+        &self.output
+    }
+}
+
+impl ConcatKdf {
+    /// The Concat KDF as defined in Section 5.8.1 of NIST.800-56A.
+    pub fn concat_kdf<D: Digest>(
```
felsweg-iota: Please add more documentation, a short explanation how the algorithm works and it's parameters.

client/src/procedures/primitives.rs
```diff
@@ -789,3 +798,115 @@ impl UseSecret for AeadDecrypt {
         &self.key
     }
 }
+
+/// Executes the concat KDF as defined in Section 5.8.1 of NIST.800-56A.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ConcatKdf {
+    /// The hash function to use in the kdf.
+    pub hash: Sha2Hash,
```
felsweg-iota, line 812: Sha-2 Family based hashes are susceptible to [length extension attacks ](https://github.com/marcelo140/length-extension). I would advise to take another digest. blake2b would be a good change. 

Or is SHA256 / X required for this algorithm? I couldn't find an explicit mentioning [here](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf)

felsweg-iota (commented):
Looks good so far, please check the inline comments :)

client/src/procedures/primitives.rs
cycraig: The SHA256 digest is specified by the JWA RFC 7518 for ECDH: https://www.rfc-editor.org/rfc/rfc7518.html#section-4.6.2

> Key derivation is performed using the Concat KDF, as defined in
>   Section 5.8.1 of [[NIST.800-56A](https://www.rfc-editor.org/rfc/rfc7518.html#ref-NIST.800-56A)], where the Digest Method is SHA-256.

client/src/procedures/primitives.rs
cycraig: The `keydatalen` is included in the optional `SuppPubInfo` parameter of Concat KDF for JWA with ECDH-ES specifically: https://www.rfc-editor.org/rfc/rfc7518.html#section-4.6.2

>    SuppPubInfo
>       This is set to the keydatalen represented as a 32-bit big-endian
>       integer.
> 
>    SuppPrivInfo
>       This is set to the empty octet sequence.

So it should not be hardcoded, and the procedure should be updated to allow the `SuppPubInfo` and `SuppPrivInfo` optional parameters per the Concat KDF specification.

cycraig:
> There are no official test vectors for the concat KDF hence why no unit test itself exists.

There is a test vector in the JWA RFC 7518 with explicit parameters and output bytes, I think we should try match that.

https://www.rfc-editor.org/rfc/rfc7518.html#appendix-C

client/src/procedures/primitives.rs
cycraig: Correct me if I'm wrong, but my understanding of the length extension attack is that:

- It requires the/an original, valid hash to perform. Here that would be the output of the Concat KDF itself, which is intended to be a shared secret and thus, if compromised, would mean the attacker would be able to decrypt any related messages anyway, even without performing a length extension attack. 
- It still results in a different hash, and thus would not compromise the derived key here.

The above does make the assumption that the output of the Concat KDF is not used as a MAC, but that seems to be the case for all intended purposes.

I agree that SHA-2 is not ideal, but given that it is specified as a requirement in RFC 7518, we still need to use it to be interoperable with existing implementations.

felsweg-iota: agreed. Looks like there is no other way around.

felsweg-iota:
> There is a test vector in the JWA RFC 7518 with explicit parameters and output bytes, I think we should try match that.
@PhilippGackstatter are you providing the impl?

PhilippGackstatter:
Thanks for the comments!

- Added documentation on the `ConcatKDF` struct which is exposed to users, and made the actual function that implements it private. Parameters are documented on the struct and no longer exist for the actual function, so there's less of a need for documentation. But let me know if the documentation is good enough, @felsweg-iota.
- Added the test vector @cycraig mentioned, thanks for finding that!
- Exposed `SuppPubInfo` and `SuppPrivInfo` as parameters, so the impl is no longer specific to how it's used in JWA.

felsweg-iota (approved):
Thanks @PhilippGackstatter for the changes. Looks good to me.

Issue #342, opened 2022-04-19
[AUDIT]: Enhance Memory Allocation For Non-Contiguous Memory Types 


felsweg-iota:
## Description
The new non-contiguous memory types provide an additional layer of security by storing allocated memory in different memory shards. The current implementation shall be enhanced to allocate memory shards very far apart, a strategy to periodically update the memory shards shall be added. 
## Motivation
Increase the security of the non-contiguous memory types.
## Requirements
Write a list of what you want this feature to do.
1. Allocate memory shards very far apart in memory space
2. Derive a strategy to periodically update the memory shards
## Open questions (optional)
None.
## Are you planning to do it yourself in a pull request?
Yes.

PR #350, opened 2022-05-02
[Security] Enhance Memory Allocation

felsweg-iota:
# Description of change
This PR adds functionality to allocate memory "far" apart with multiple strategies. 

#342 

## Type of change
The feature has been tested with local unit tests

## Change checklist


engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+use std::{fmt::Debug, ptr::NonNull};
+
+/// Fragmenting strategy to allocate memory at random addresses.
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum FragStrategy {
+    /// Anonymously maps a region of memory
+    Map,
+
+    /// System's allocator will be called a few times
```
vuongDang: I find this comment confusing. The idea is to "fragment by allocating empty region of memory" right?

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+        let mut rng = thread_rng();
+
+        let pagesize = nix::unistd::sysconf(nix::unistd::SysconfVar::PAGE_SIZE)
+            .unwrap_or(Some(0x1000i64))
+            .unwrap() as usize;
+        info!("Using page size {}", pagesize);
+
+        unsafe {
+            loop {
+                let mut addr: usize = (0x00007fff00000000 | (rng.gen::<usize>() >> 32)) & (!0usize ^ (pagesize - 1));
```
vuongDang: This needs a comment. 
Why 0x00007fff00000000? is it the start address of the memory region?

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+                info!("preallocated segment {:p}", ptr);
+
+                if ptr == libc::MAP_FAILED {
+                    continue;
+                }
+
+                // write random bytes into allocated pages
+                let bytes = ptr as *mut usize;
+                let end = rng.gen_range(size..(size * 0x1000));
+                (size..end).for_each(|_| bytes.write(rng.gen()));
```
vuongDang: I may be wrong, but doesn't this just rewrite random value at the same location repeatedly?

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+
+                info!("preallocated segment {:p}", ptr);
+
+                if ptr == libc::MAP_FAILED {
+                    continue;
+                }
+
+                // write random bytes into allocated pages
+                let bytes = ptr as *mut usize;
+                let end = rng.gen_range(size..(size * 0x1000));
```
vuongDang: How much memory do we want to fill with random data? 
I'd say one page or `desired_alloc_size`. I don't get the value `end` here

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+                (size..end).for_each(|_| bytes.write(rng.gen()));
+
+                #[cfg(any(target_os = "macos"))]
+                {
+                    // on linux this isn't required to commit memory
+                    let error = libc::madvise(&mut addr as *mut usize as *mut libc::c_void, size, libc::MADV_WILLNEED);
```
vuongDang: Can you give me source of `madvise` necessity in macos, please? I've never about this peculiarity

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    #[cfg(target_os = "windows")]
+    fn alloc() -> Result<NonNull<T>, Self::Error> {
+        use random::{thread_rng, Rng};
+        let mut rng = thread_rng();
+
+        let handle = windows::Win32::Foundation::INVALID_HANDLE_VALUE;
+
+        unsafe {
+            // allocation prelude
```
vuongDang: Why do we need an allocation prelude? Sorry I don't know anything about Windows

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+        let max = 0xFFFF_FFFF;
+
+        #[cfg(any(target_os = "unix", target_os = "linux"))]
+        let _pagesize = nix::unistd::sysconf(nix::unistd::SysconfVar::PAGE_SIZE)
+            .unwrap_or(Some(0x1000i64))
+            .unwrap() as usize;
+
+        // Within the loop we allocate a sufficiently "large" chunk of memory. A random
+        // offset will be added to the returned pointer and the object will be written. This
+        // actually leaks memory.
```
vuongDang: I thought the idea for non contigous memory was: 
1. malloc fragment1 
2. malloc big chunk  
3. malloc fragment2 
4. free the big chunk to save memory

I guess your strategy allows you to allocate a single piece of memory hidden with a big chunk, but the downsides is that you actually use a lot of memory.

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,427 @@
...
+                let actual_mem = ((mem_ptr as usize) + offset) as *mut T;
+                actual_mem.write(T::default());
+
+                return Ok(NonNull::new_unchecked(actual_mem));
+            }
+        }
+    }
+
+    #[cfg(target_os = "windows")]
+    fn alloc() -> Result<NonNull<T>, Self::Error> {
```
vuongDang: Isn't this a normal allocation?

vuongDang (commented):
Nice pr and nice job looking into the memory allocations of the different architectures!

Just wondering about 2 points: 
- Do we need to fill the memory chunks with random data? I think this is costly. In the use case of non-contiguous memory, with the _mmap_ strategy, jsut allocating 2 random mmap areas seems more efficient to me 
  
- More or less the same question with malloc strategy. In the use case of non-contiguous memory allocating a big chunk of empty memory feels more efficient

Are we targetting another use case that non-contiguous memory?

engine/runtime/src/memories/frag.rs
felsweg-iota: I totally agree with you. That "magic" constant needs more explanation. Some source claimed, that this is the beginning of memory allocations in user space.

engine/runtime/src/memories/frag.rs
felsweg-iota: True :sweat_smile:  The `bytes` pointer is not being advanced.

engine/runtime/src/memories/frag.rs
felsweg-iota: Good catch. That's actually a bug. `0x1000` would be the (default)  page size (4096), but `DirectAlloc` makes an api call to retrieve the actual number.

engine/runtime/src/memories/frag.rs
felsweg-iota: The prelude is a result of some testing under Windows (10). It looks like, that the windows API requires another mapped chunk to advance an internal pointer.

engine/runtime/src/memories/frag.rs
felsweg-iota: ~It is! And Windows obviously allocates successive chunks of memory at random places. At least CI and my local machine with Windows seem to agree.~

It is! The only difference is, that a random address is given to create an anonymous mapping at the random location.

engine/runtime/src/memories/frag.rs
felsweg-iota: Which one are you referring to? Otherwise I would say the `FragStrategry` is exactly that: `allocate memory at random addresses`. Maybe an addendum is missing. Something like "[...] that are reasonably far apart" Even though "reasonably" cannot be expressed numerically :shrug:

engine/runtime/src/memories/frag.rs
felsweg-iota: I agree on the theoretical approach, but macos seems to behave differently. A previous approach similar to yours showed that macos does not not enforce initializing the memory, if another call to `malloc` followed. The current setup is very inefficient and leaks memory. Additionally it's hard to actually test low-level allocation on a mac without having one, but relying on the CI tests return positive.

vuongDang: What do you mean by "does not enforce initializing the memory"? 
Is the big chunk at step 2. do not get allocated at all or just does not get filled with random data?

engine/runtime/src/memories/frag.rs
vuongDang: "System's allocator will be called a few times". 
And now that I've read the pr the `Direct` strategy doesn't do that currently. Now it allocates a big chunk and hide the relevant memory within

felsweg-iota: what do you mean? The "algorithm" loops over the `malloc` and calls it again, if it fails.

engine/runtime/src/memories/frag.rs
felsweg-iota: My observation was that allocating a "huge" chunk of memory gets allocated, but discarded after a second call eg. the address returned was close by, when allocating another chunk. I would have expected to have the second chunk nowhere "near" the last one, but I can supply a test to re-confirm.

vuongDang: ah ok I get it, is it possible that the big chunk got allocated somewhere else? I know that some implementation of `malloc` use memory buckets: memory is pre-allocated and ordered by size, depending on the size you require you get memory from different buckets

felsweg-iota: > ah ok I get it, is it possible that the big chunk got allocated somewhere else

Would you have a strategy for that?

tensor-programming:
It looks like the windows API needs some minor tweaks to be more stable. 

I ran the engine benchmark on my machine (windows 11) and it blacked out my screen when it hit the `allocate memory direct` iteration.  It killed mattermost, explorer, chrome, vscode and a few other programs and I had to restart the machine entirely.  Ill take a look closer at the Windows API itself to see if there is any specific reason.  I might have just been a result of the benchmark running so many iterations but if thats the case, it means larger strongholds will be suspectable to this kind of behavior.

felsweg-iota:
@tensor-programming this indeed sounds like a serious issue. I suspect the lack of [deallocation on the failed condition](https://github.com/iotaledger/stronghold.rs/blob/d5fc82296b89dc3a353480a98022f026137666aa/engine/runtime/src/memories/frag.rs#L432) to be the cause. But I would have to verify.

engine/runtime/README.md
```diff
@@ -83,3 +83,8 @@ Hence data security depends on the strength of the encryption scheme and the 'ob
     - [ ] access to the locked memory
 - [ ] Benchmarks
 - [ ] no-std
+
+
+- 
+0x00007ffff7fc4fe0
```
felsweg-iota: better remove the magic constants from the readme

engine/runtime/src/boxed.rs
```diff
@@ -145,10 +146,40 @@ impl<T: Bytes> Boxed<T> {
         if unsafe { sodium_init() == -1 } {
             panic!("Failed to initialize libsodium")
         }
+        let mut rng = random::thread_rng();
+
+        // TESTING
```
felsweg-iota: Are the comments part of the "virtual" allocation of more fragments, to indicate a block with some "extra work" eg. allocate some data to create space between the shards?

engine/runtime/src/boxed.rs
```diff
@@ -145,10 +146,40 @@ impl<T: Bytes> Boxed<T> {
...
+        let chunk_a = unsafe {
+            let size = rng.gen_range(0x1000..0xFFFFFF);
```
felsweg-iota: Maybe we can put this into a constant, if useful.

engine/runtime/src/boxed.rs
```diff
@@ -205,6 +236,13 @@ impl<T: Bytes> Boxed<T> {
     fn is_locked(&self) -> bool {
         self.prot.get() == Prot::NoAccess
     }
+
+    #[cfg(test)]
+    #[allow(dead_code)]
+    /// Returns the address of the pointer to the data
+    pub fn get_ptr_address(&self) -> usize {
```
felsweg-iota, line 212: Is this still being used? Also may be better to use `std::ptr::addr_of!` macro.

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+/// Fragmenting strategy to allocate memory at random addresses.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum FragStrategy {
+    /// Anonymously maps a region of memory
+    Map,
+
+    /// Using system allocator (`malloc` on linux/bsd/macos and `VirtualAlloc` on windows)
+    Direct,
+
+    /// Allocate using both strategy
```
felsweg-iota:
```suggestion
    /// Allocate using both strategies
```

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+//! - Memory Mapped: anonymous memory is being mapping, the memory address will be randomly selected.
+
+use crate::MemoryError;
+use log::*;
+use std::{fmt::Debug, ptr::NonNull};
+use zeroize::Zeroize;
+
+// The minimum distance we consider between 2 fragments
+// This is the page size for most linux systems
+pub const FRAG_MIN_DISTANCE: usize = 0x1000;
```
felsweg-iota: yeah, that's linux "default" page size, I wonder if we should keep this somewhat externally configurable :thinking:

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+    /// Returns a fragmenting allocator by strategy
+    ///
+    /// # Example
+    ///
+    /// ```skip
+    /// use stronghold_engine::runtime::memories::*;
+    ///
+    /// let object  = Frag::by_strategy(FragStrategy::Default).unwrap();
+    /// ```
+
```
felsweg-iota: nitpicking: empty line

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+            )));
+        }
+
+        info!("Mapped 2 fragments: at {:?} and {:?}", a.ptr, b.ptr);
+        Ok((a, b))
+    }
+
+    /// Tries to allocate two objects of the same type with a default minimum distance in memory space of
+    /// `FRAG_MIN_DISTANCE`.
+    pub fn alloc(strategy: FragStrategy, data1: T, data2: T) -> Result<(Frag<T>, Frag<T>), MemoryError> {
```
felsweg-iota: The naming `alloc2` is a remnant of the past and returned two fragments, while `alloc` returned one. I would rename the functions to

- `alloc` -> `alloc_initialized`
- `alloc2` -> `alloc` or `alloc_default`

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+            FragStrategy::Direct => DirectAlloc::alloc(config),
+            FragStrategy::Map => MemoryMapAlloc::alloc(config),
+            FragStrategy::Hybrid => MemoryMapAlloc::alloc(config),
+        }?;
+
+        let actual_distance = calc_distance(a.get()?, b.get()?);
+        if actual_distance < distance {
+            error!(
+                "Distance between parts below threshold: \na: 0x{:016x} \nb: 0x{:016x} \ngiven_value: 0x{:016x}",
+                a.ptr.as_ptr() as usize,
```
felsweg-iota: you can use `std::ptr::addr_of!` here

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    fn dealloc(frag: &mut Frag<T>) -> Result<(), Self::Error> {
+        dealloc_map(frag.info.0, frag.info.1 as libc::size_t)
+    }
+
+    #[cfg(target_os = "windows")]
+    fn alloc(_config: Option<FragConfig>) -> Result<Frag<T>, Self::Error> {
+        // use random::thread_rng;
```
felsweg-iota:
```suggestion
```

engine/runtime/src/memories/frag.rs
```diff
@@ -0,0 +1,624 @@
...
+        // let mut rng = thread_rng();
```
felsweg-iota:
```suggestion
```

engine/runtime/tests/memory_leak_test.rs
```diff
@@ -0,0 +1,98 @@
...
+        Self {
+            id: 123456789,
+            name: "Some heap allocated value".to_owned(),
+        }
+    }
+}
+
+const NB_ALLOC: usize = 20;
+// We run this the test binary with valgrind to check for potential leak
+#[test]
```
felsweg-iota, line 28: please mark with `#[ignore]`

felsweg-iota (commented):
@vuongDang Looks good to me. 

I like the approach to provide an optional config to check for previous allocations.

engine/runtime/src/boxed.rs
vuongDang: I didn't notice that this code was added. I guess this is used to create space between the shards. But as it is, this is used even for normal memory allocation,  I will check it it is necessary

engine/runtime/src/memories/frag.rs
vuongDang: I can make it a static variable that is initialized once with system page size or with a default value if we can't get it

engine/runtime/src/boxed.rs
vuongDang: This is used for testing. also `addr_of` won't give the exact address of where the data is stored since this is encapsulated into multiple structure with private fileds

vuongDang (approved):

Issue #339, opened 2022-04-14
[Request] Add a procedure to concatenate secrets


PhilippGackstatter:
## Description

Add a procedure that concatenates two secrets. Specifically, the functionality requested is this one: https://doc.rust-lang.org/std/primitive.slice.html#method.concat

## Motivation

We need this for the `ECDH-1PU` algorithm that concatenates the secrets generated from two Diffie-Hellman procedures: https://github.com/iotaledger/identity.rs/blob/dd55652ccec75cbe3f095b162e83e8efec6b08ec/libjose/src/jwe/encoder.rs#L701.

See also the relevant section from the RFC: https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2.3

## Requirements

Write a list of what you want this feature to do.

- concatenate two secrets

## Open questions (optional)

n/a

## Are you planning to do it yourself in a pull request?

I'm open to it, just let me know.

felsweg-iota:
Hey @PhilippGackstatter, I think this Issue has been resolved by [354](https://github.com/iotaledger/stronghold.rs/pull/354). May you confirm?

PhilippGackstatter:
Unfortunately not. The `ConcatKdf` does something different. This issue really just needs to expose the [`concat`](https://doc.rust-lang.org/std/primitive.slice.html#method.concat) method on slices as a procedure.

Issue #338, opened 2022-04-14
[Request] Support AES key wrapping


cycraig:
## Description

Add support for key wrapping algorithms such as AES: A128KW, A192KW, A256KW.
https://datatracker.ietf.org/doc/html/rfc3394

These are available under the `aes-kw` feature in the iota-crypto crate: https://github.com/iotaledger/crypto.rs/blob/dev/src/ciphers/aes_kw.rs

## Motivation

The [ECDH-1PU](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#page-4) algorithm for encrypting data using secrets derived by eliptic-curve Diffie-Hellman has two "modes". A "Direct" mode which uses a concatenated derived secret as the content encryption key directly, and a "Key Wrapping" mode which applies one of A128KW, A192KW, or A256KW to the concatenated secret, using the result as the content encryption key.

See the following RFCs for reference:
- https://datatracker.ietf.org/doc/html/rfc7518#section-4.6
- https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2

~As far as I am aware, it is currently only possible to use the "Direct" mode with Stronghold's exposed procedures for Diffie-Hellman and AES256 encryption.~ Edit: without secret concatenation it doesn't look like we can implement the "Direct" mode either.

The IOTA Identity team requires the above-mentioned key wrapping algorithms to satisfy the DIDComm specification when using Stronghold, which requires A256KW specifically for X25519 keys: https://identity.foundation/didcomm-messaging/spec/#key-wrapping-algorithms

## Requirements

Write a list of what you want this feature to do.

1. Expose a procedure or mode to use A128KW, A192KW, or A256KW to wrap a derived secret for encryption/decryption.

## Open questions (optional)

- Unsure if key wrapping should be applied as an independent procedure to derive a key, similar to Slip10, or as an option passed to the Diffie-Hellman or encryption/decryption procedures. Having an independent procedure may be better for separation of concerns but require temporary secret locations that need to be cleared up afterwards?
  - Apparently chained procedures make temporary secrets from intermediate procedures less of an issue.
- ~Unsure if more key concatenation procedures are required to satisfy ECDH-1PU.~ 
  - We also require support for secret concatenation KDF. See #339.

## Are you planning to do it yourself in a pull request?

If you want me to.

felsweg-iota:
@cycraig @PhilippGackstatter  Can this issued be closed?

[...]

PR #354, opened 2022-05-07
 Allow procedures to use multiple secrets 

PhilippGackstatter:
# Description of change

Allow procedures to use multiple secrets, which is needed for #338 and #339.

A careful review of the synchronization code (when things are locked, and in what manner (write, read)) would be appreciated. 

A downside of using const generics is that one cannot easily convert `[T;N]` into `[X;N]`, because when using iterators, one loses the size information and so we cannot easily `collect` into such a new array. The workaround is to collect into a `Vec<X>` and use [`<[_; N]>::try_from(vec).expect("...")`](https://doc.rust-lang.org/std/primitive.array.html#method.try_from-4) for conversion, which [does not appear to clone buffers unnecessarily](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3052).

Open Question: Should we re-add more of the procedure tests that once existed? Currently, the state of procedure testing is very minimal, unless I've missed them somewhere?

n/a

## Type of change

- [ ] Documentation Fix

#355 uses `UseSecret<2>` and works as intended.

Existing tests for modified procedures pass locally.

## Change checklist

### DIFF
--- a/client/src/procedures/clientrunner.rs
+++ b/client/src/procedures/clientrunner.rs
@@ -185,7 +185,8 @@ impl Client {
             let key: Key<Provider> = keystore.get_key(vault_id).ok_or(VaultError::VaultNotFound(vault_id))?;
             ids.push((key, vault_id, record_id));
         }
-        let ids: [(Key<Provider>, VaultId, RecordId); N] = ids.try_into().expect("buffers did not have exactly len N");
+        let ids: [(Key<Provider>, VaultId, RecordId); N] =
+            <[_; N]>::try_from(ids).expect("buffers did not have exactly len N");
         Ok(ids)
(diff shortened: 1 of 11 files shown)


felsweg-iota:
> Open Question: Should we re-add more of the procedure tests that once existed? Currently, the state of procedure testing is very minimal, unless I've missed them somewhere?

@PhilippGackstatter yes. I will re-add the existing procedures tests from the old branch.

felsweg-iota (approved):
Thank you @PhilippGackstatter for the work done. Looks good from my point of view. I still wonder, if the procedures API can be further improved to be more flexible / enable more composable structures regarding the number of locations being used.

PR #355, opened 2022-05-07
Expose procedures for AES Key Wrap

### LINKED ISSUE #338: [Request] Support AES key wrapping

PhilippGackstatter:
# Description of change

Adds `AesKeyWrapEncrypt` and `AesKeyWrapDecrypt` procedures using Aes256, with the option to add Aes128 and Aes192 non-breakingly in the future.

## Encryption

This takes two locations as input, `encryption_key` and `wrap_key`. `encryption_key` will be used to encrypt `wrap_key`, and return the resulting ciphertext bytes to the caller. This looks suspiciously like leaking secrets, but in reality it doesn't. Without the `encryption_key`, no one is able to get the encrypted `wrap_key` secret, except for other parties that can also produce the `encryption_key`, e.g. if this key was created through a Diffie-Hellman operation.

This is exactly the use case we need it for: JSON Web Encryption. We have some secret `verification_method` key material lying in Stronghold, and presumably, so does the peer we want to communicate with. Through DH we can produce a shared secret - this is the `encryption_key` in our case. We also randomly generate another secret key for content encryption, which is our key to be wrapped (the `wrap_key`). Now we wrap the `wrap_key` with `encryption_key` and add it to the message. The receiver can do DH, decrypt the `wrap_key` and decrypt the content. We don't reuse the content encryption key (`wrap_key`).

In general, even if someone leaks a secret in encrypted form, in order for it to be useful someone would still need to somehow gain access to the `encryption_key`, which is not possible through stronghold. All in all, this doesn't seem abusable to me, even by users willing to, but I'd be happy if someone else validates this.

## Decryption

This takes the location of the `encryption_key` and the `wrapped_key` as input, in form of a `Vec<u8>`. The key is unwrapped and written into a location in a vault.

closes #338

This feature required procedures to be able to take two secrets as input, which required some refactoring: #354.

## Type of change

- [ ] Documentation Fix

Added a roundtrip test based on an official test vector which is also used in crypto.rs.

## Change checklist

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -20,7 +20,7 @@ std                     = []
 thiserror               = { version = "1.0.30" }
 zeroize                 = { version = "1.4.3", features = ["zeroize_derive"] }
 serde                   = { version = "1.0",   features = [ "derive" ] }
-iota-crypto             = { version = "0.8.0", features = [ "aes", "random", "ed25519", "sha", "hmac", "bip39-en", "bip39-jp", "slip10", "chacha", "x25519" ] }
+iota-crypto             = { version = "0.8.0", features = [ "aes", "aes-kw", "random", "ed25519", "sha", "hmac", "bip39-en", "bip39-jp", "slip10", "chacha", "x25519" ] }
 hkdf                    = { version = "0.11" }
 bincode                 = { version = "1.3" }
 pin-project             = { version = "1.0.10", optional = true }
--- a/client/src/procedures.rs
+++ b/client/src/procedures.rs
@@ -8,9 +8,10 @@ mod types;
 pub use clientrunner::*;
 
 pub use primitives::{
-    AeadCipher, AeadDecrypt, AeadEncrypt, BIP39Generate, BIP39Recover, Chain, ChainCode, ConcatKdf, CopyRecord,
-    Ed25519Sign, GarbageCollect, GenerateKey, Hkdf, Hmac, KeyType, MnemonicLanguage, Pbkdf2Hmac, PublicKey, RevokeData,
-    Sha2Hash, Slip10Derive, Slip10DeriveInput, Slip10Generate, StrongholdProcedure, WriteVault, X25519DiffieHellman,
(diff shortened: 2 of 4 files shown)


felsweg-iota (approved):
Looks good to me

PR #369, opened 2022-06-02
Fix examples path + typo

r-c-k:


felsweg-iota (approved):
@r-c-k Thanks for fix!

PR #383, opened 2022-06-20
Feat/software transactional memory

felsweg-iota:
# Description of change

This PR implements a variation of the TL2 algorithm described by Shavit et al. Access to shared memory is being locked by a specialized bounded spin-lock with integrated versioning. The algorithm differentiates between reading and writing transaction, while having some reading transaction performance optimization in place.
none
## Type of change
Unit checks have been provided to verify functional correctness

## Change checklist


rlu/src/stm/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+    /// this writes the value into the transactional log
+    pub fn store(&mut self, tvar: &TVar<T>, value: T) -> Result<(), TxError> {
+        self.write.insert(tvar.clone(), value);
+
+        Ok(())
+    }
+
+    #[inline(always)]
+    fn lock_write_set(&self) -> Result<(), TxError> {
+        let fail_lock_write = AtomicBool::new(false);
```
vuongDang: Do we need an atomic here? Each transaction is handled sequentially

rlu/src/stm/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+
+        for tvar in self.write.keys() {
+            info!("TRANSACTION({:?}): WRITE LOCK", self.id);
+
+            if tvar.lock.try_lock().is_err() {
```
vuongDang: The current implementation of `try_lock()` only returns `Ok(_)`

rlu/src/stm/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+            }
+
+            // lock to thread id
+            tvar.lock(self.id)?;
+        }
+
+        if fail_lock_write.load(std::sync::atomic::Ordering::SeqCst) {
+            for tvar in self.write.keys() {
+                tvar.lock.unlock()?;
+            }
```
vuongDang: Some tvar of the write set may be locked by another thread (since we failed to get all the write locks).
Shouldn't we only release the tvar we managed to get a lock for?

rlu/src/stm/mod.rs
```diff
@@ -0,0 +1,659 @@
...
+    /// Validates the read set
+    #[inline(always)]
+    fn validate(&self, wv: usize) -> Result<(), TxError> {
+        let rv = self.version;
+        if rv + 1 == wv {
+            return Ok(());
+        }
+
+        for tvar in &self.read {
+            if tvar.lock.version() >= rv {
```
vuongDang: I believe it's `tvar.lock.version() > rv`, the versions may be equal

rlu/src/stm/version.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// Tries to acquire a lock and returns an `Ok(())` on success.
+    ///
+    /// # Example
+    /// ```
+    /// use stronghold_rlu::stm::VersionLock;
+    /// let lock = VersionLock::default();
+    /// lock.try_lock().expect("Failed to acquire lock");
+    /// assert!(lock.is_locked());
+    /// ```
+    pub fn try_lock(&self) -> Result<(), TxError> {
```
vuongDang: This spinlock is not bounded. My guess is that we can have a deadlock when two threads try to lock the tvar of their write set at the same time. That's why the paper asks for bounded spinlock

felsweg-iota: True. The question is, if a bounded spinlock needs some empirical value to actually make the algorithm work. The bounded version is currently commented out, but would offer a very high bound of runs. Using sleep reduces the CPU load.

rlu/src/stm/mod.rs
felsweg-iota: This relates to your comment above, since try-lock is unbounded, there is no way to return an `Err`, but `Ok`.

rlu/src/stm/mod.rs
felsweg-iota: true, we can skip the atomic here

rlu/src/stm/mod.rs
felsweg-iota: hmm the `TVar` here would be a pointer a one managed transactable memory (eg. the HashSet from the test ). Each `Transaction` intending to write the changes to memory needs to lock all `TVars` (eg. references to this one single transactable memory ) in order to be consistent. Since a `TVar` itself contains only atomic references to the actual version and memory, locking one `TVar` locks them all, but by a specific Transaction.

vuongDang:
This has been migrated to PR #441

PR #392, opened 2022-06-27
doc: skeleton doc concepts files

felsweg-iota:
# Description of change

PR for key concepts explained for the wiki

none

## Type of change

- [x] Documentation Fix

it's documentation.

## Change checklist


documentation/docs/key_concepts/non-contiguous-data-types.md
```diff
@@ -0,0 +1,84 @@
...
+-->
+
+# Non-Contigouos Data Types and Handling secrets at Runtime
+
+#### Authors: Matthias Kandora - \<matthias.kandora@iota.org>
+
+***Abstract:***
+
+
+Running processes store objects in allocated memory contiguously, meaning the stream of bytes is consecutive. This is not always desirable, as sensitive information can be easily read out from parts of the memory. Here we describe what non-contigouos memory data structures are and how they work.
```
lucas-tortora:
```suggestion
# Non-Contiguous Data Types and Handling secrets at runtime

Running processes store objects in allocated memory contiguously, meaning the stream of bytes is consecutive. This is not always desirable, as an attacker could easily read sensitive information from parts of the memory. This section will describe non-contiguous memory data structures and how they work.
```

documentation/docs/key_concepts/non-contiguous-data-types.md
```diff
@@ -0,0 +1,84 @@
...
+#### Authors: Matthias Kandora - \<matthias.kandora@iota.org>
+
+***Abstract:***
+
+
+Running processes store objects in allocated memory contiguously, meaning the stream of bytes is consecutive. This is not always desirable, as sensitive information can be easily read out from parts of the memory. Here we describe what non-contigouos memory data structures are and how they work. 
+
+***On Memory Management*** 
+
+Whenever a computer program requires memory to store data (eg. some text, an image, etc. ) it needs to allocate memory. By "allocating memory" we refer to a mechanism, which is given by the operating system to the running program to ensure, that some of the memory is given to the program for some purpose. In this context we also speak of "virtual memory space", as each program running ( or in computer science parlance "processs" ) does not need to know the "real" address in "real" memory, but only an abstracted / virtualized version of it.
```
lucas-tortora:
```suggestion
## Memory Management

Whenever a computer program requires memory to store data (e.g., some text, an image, etc. ), it must allocate memory. Allocating memory is the mechanism by which the operating system grants some of its storage and processing power to a running program to ensure that the problem can run properly.

In this context, there is also what is commonly known as "virtual memory space". Every program (process) running does not need to know the actual address in the memory but only an abstracted version.

The Stronghold runtime provides memory (speak managed allocation) types to protect sensitive data.
```

documentation/docs/key_concepts/non-contiguous-data-types.md
```diff
@@ -0,0 +1,84 @@
...
+
+The Stronghold runtime provides memory ( speak managed allocation ) types for handling sensitive data. One such type is `Boxed`. This type locks allocated memory and prevents it from being recorded in a memory dump. Since locking memory is dependent on the operating system, the `Boxed` type relies on libsodium `sodium_mlock` function, which calls the `mlock` function on linux or equivalent functions on other operating systems. `mlock` prevents the current virtual address space of the process to be paged into a swap area, thus preventing the leakage of sensitive data. This in turn will be used by guarded heap allocations of memory.
+
+Guarded heap allocations work by placing a guard page in front and at the end of the locked memory, as well as a canary value at the front. The schematic view visualizes it. 
+
+![](https://i.imgur.com/oy0Ri1Z.png)
```
lucas-tortora:
```suggestion
### Boxed Memory

`Boxed` memory locks allocated memory and prevents it from being recorded in a memory dump. Since locking memory is dependent on the operating system, the `Boxed` type relies on [Libsodium’s](https://libsodium.gitbook.io/doc/) `sodium_mlock` function. This function calls the `mlock` function on Linux (or equivalent functions on other operating systems). `mlock` prevents the current virtual address space of the process from being paged into a swap area, preventing the leakage of sensitive data. This, in turn, will be used by guarded heap allocations of memory.

Guarded heap allocations work by placing a guard page in front and at the end of the locked memory and a canary value at the front. The schematic view visualizes it.

![Guarded Heap Memory Allocations](https://i.imgur.com/oy0Ri1Z.png)
```

documentation/docs/key_concepts/non-contiguous-data-types.md
```diff
@@ -0,0 +1,84 @@
...
+ fn unlock(&self) -> Result<Buffer<u8>, MemoryError>;
+```
+
+Currently, the trait is implemented by three types of memory:
+
+| Type                  | Description                                                                                                   |
+|:----------------------|:--------------------------------------------------------------------------------------------------------------|
+| `RamMemory`           | The allocated value resides inside the system's ram.                                                          |
+| `FileMemory`          | The allocated value resides on file system                                                                    |
+| `NonContiguousMemory` | Allocated memory is being fragmented either across the system's ram or file system, or a combination of both. |
```
lucas-tortora:
````suggestion
Libsodium provides three types to guard memory:

| function                    | description                                                                           |
|:----------------------------|:--------------------------------------------------------------------------------------|
| `sodium_mprotect_noaccess`  | Makes the protected memory inaccessible. It can neither be read from nor written to. |
| `sodium_mprotect_readonly`  | Makes the protected memory read-only. Memory can be read from but not written to.     |
| `sodium_mprotect_readwrite` | Enables reading from and writing to protected memory.                                 |


Stronghold exposes locked memory via the `LockedMemory` trait, that exposes two functions that need to be implemented:

```rust
/// Modifies the value and potentially reallocates the data.
fn update(self, payload: Buffer<u8>, size: usize) -> Result<Self, MemoryError>;

/// Unlocks the memory and returns a Buffer
fn unlock(&self) -> Result<Buffer<u8>, MemoryError>;
```

Currently, three types of memory implement this trait:

| Type                  | Description                                                                                      |
|:----------------------|:-------------------------------------------------------------------------------------------------|
| `RamMemory`           | Allocated values reside inside the system's ram.                                                 |
| `FileMemory`          | Allocated values reside on the file system                                                       |
| `NonContiguousMemory` | Allocated memory is fragmented across the system's ram or file system, or a combination of both. |

````

File: documentation/docs/key_concepts/non-contiguous-data-types.md [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+
+*** Conclusion ***
+
+With the new runtime we now have quite a few options to ensure sensitive data in memory is protected. Non-contiguous types are fairly new to Stronghold and we have to figure out what is a good balance between performance ( everything is in RAM ) and security ( fragmented across RAM and file system). There is another limit put by the operating system, when it comes to the maximum number of protected memory regions. On some Linux machines we encountered empirically that the limit was about ~8000 guarded pages. In order to fix that, we decided to not store guarded pages inside a vault, but guard it on demand. The number of sensitive data at rest is presumably higher compared to sensitive data being present for cryptographic procedures. Sensitive entries inside the vault at rest are encrypted with XChaCha20-Poly1305, which gives us some decent security.
```
lucas-tortora:
```suggestion
## Non-Contiguous Data Types

Under normal circumstances, the allocated memory is continuous and page-aligned. The operating system provides memory blocks of a minimum predetermined size. Data types that do not have a multiple of some minimum number in bytes are padded with zeroes. Metadata describes the actual fields. Operating systems take this approach to improve performance as loading some larger chunks of 2^n bytes is faster than loading the exact number of bytes.

Non-contiguous (NC) data types store their inner referenced data in multiple locations, either in memory, on the file system, or a mixture of both. NC data types are useful if the memory is partitioned into multiple segments, and storing a continuous stream of bytes might not be possible. The disadvantage is that the operating system must constantly keep track of the referenced memory segments.


### Boojum Scheme

Non-contiguous memory types split protected memory into multiple fragments, mitigating any memory dumps and making it virtually impossible for attackers to retrieve stored data. The following section describes non-contiguous memory types in more detail with a use case we often encountered and solved when we were developing Stronghold.

#### Use Case: Passphrase Management

Proper passphrase management was one of the most challenging tasks during the development of Stronghold. You need a password whenever you want to load a persistent state from a snapshot file. If you were the only user of Stronghold, and reading and writing would be interactive, providing the password each time would not be a problem. The time window in which you would use the passphrase to decrypt and later encrypt to persist a state would be small and almost non-predictable.

However, consider an application that requires constant writing into a snapshot, meaning the passphrase to encrypt the snapshot must be stored in memory. If an attacker gains access to the machine, they could dump the memory of the running process and read out the passphrase in plaintext, which of course, would be a significant security problem. Luckily, there is a solution to that called the Boojum Scheme, as described by Bruce Schneier et al. in “Cryptography Engineering”.

## Conclusion

With the new runtime, Stronghold has several options to protect sensitive data in memory. Non-contiguous types are fairly new to Stronghold. We have to figure out a good balance between performance (everything is stored in RAM) and security (fragmented across RAM and file system).

There is another limiting factor. Regarding the maximum number of protected memory regions, we empirically encountered the limit of about 8000 guarded pages on some Linux machines. To fix that, we decided to avoid storing pages inside a vault and guard them on demand instead. The amount of sensitive data at rest is presumably higher compared to sensitive data present for cryptographic procedures. Sensitive entries inside the vault at rest are encrypted with XChaCha20-Poly1305, which provides security while circumventing this limitation.
```

documentation/docs/key_concepts/procedures.md
```diff
@@ -0,0 +1,73 @@
...
+// used to check for errors
+let procedure_result = client.execute_procedure(StrongholdProcedure::GenerateKey(generate_key_procedure));
+ 
+// Front the previously generate key, we want to export the public key
+let public_key_procedure = stronghold::procedures::PublicKey {
+       ty: keytype,
+       private_key: output_location,
+};
+```
+
```
lucas-tortora:
````suggestion
# Cryptographic Procedures

Stronghold ensures that sensitive data cannot easily escape from memory. This requires a mechanism to work with secrets stored inside Stronghold's vault. While parts of Stronghold are primarily concerned with writing secrets, the question arises: "What can you do with a secret that is never exposed? Why even store it?”.

## Cryptographic Procedures Pipeline

Stronghold features a framework to build pipelines of cryptographic operations. The pipeline pattern is an abstraction over chained function calls. Each stage of a pipeline can either:

* Produce a value. For example, generate a secret (BIP39 and Mnemonic, Ed25519)
* Process an existing value. For example, deriving a secret key from an existing key (SLIP10)
* Export a value from an existing secret. For example, export the public key of a key pair.

The framework is abstracted in a way that allows you to combine simple and complex cryptographic procedures. However, custom procedures are not possible within Stronghold. This is because a procedure can access secrets. Providing a custom procedure that exposes a secret and returns it would violate the Stronghold core principle.

The procedures framework is built upon a pipeline pattern. Each stage is only given the location inside the vault to access or work with secret data. The following schematic showcases the generation of a public / private keypair that gets stored in `location1`. The following step takes the previous `location1` and derives a new keypair to store it in `location2`. Eventually, the last stage takes `location2`, extracts the public key and returns it to some publicly accessible data.

[![Generation of a Public / Private Keypair](https://i.imgur.com/qZ5QX23.png)](https://viewer.diagrams.net/?tags=%7B%7D&highlight=FFFFFF&edit=_blank&layers=1&nav=1&title=pipeline#R7Vldb6M4FP01eZzI2JDAY9t0u1JnpEp92Jl9c8ED3nEw45gm7K%2Ff62K%2BG5o2JO1oFlUq9%2Fr6g3PPPQZnRq7WuxtFs%2BSLjJiYYRTtZmQ1w9hxMZ6ZPxQVpWcZkNIRKx7ZoMZxz%2F9l1omsN%2BcR23QCtZRC86zrDGWaslB3fFQpue2GfZeiO2tGYzZw3IdUDL1K2F8Ugnpdf3UOP%2Fk%2FE4qWZ2kG1Z0yrYOjYJjeS25SLXM3KlpNTl3Xp3xYQBr8Kl7PfHntZ6YYql%2BpAOJL1lwrveZTcX%2BGex8i7p37efHJuNRypy%2B8QzvBAw4GXEHw2GgsfpU8PiZ26WeqmenrY24S42%2Fx9pLnTVF1bx1L1sss%2BviwpUJfM0YmZdCJq3CdfsPqOhad0CjcCX6LUAy4HbagXlvDARU5pDgi6s%2F0FqLdfQsNFK%2FqizRMDznQtxJYVUYKcyZfVa2qhZIM2wbNdyWRRvmFwzrQoIsa0Lm1DL6NreNvzwKhYkbW5UgdRyMq6HbtIGNzZzr8nicjyLZRoyJUMW5QrKCSOWxhwAsVEPqsnV9Bl8Jg0HJ3X6fDkH5wudLF%2FBAFAWgexYUyqdyFimVFw33ssGcoNpE%2FNZysw6%2F2FaF1ZDaa5lNw2AoCq%2Bto1vJn9zrzJXO5vP0ioqa8f116dI7FnzWzUK3De9jFF12pu3jcxVyMbAsSpPVcz0WJxfBhrkRmmgmKCaP3YF%2Fbmc2q53ksOaa%2FrUNKjq3XW7Q5RPZHs1zLhQihatsMwEbEbmcXrz%2BD09f1083JQraGhaY%2FJ25uLFUcxFRzIXvY65pyehNzUHj9OVwTZww1KmqAahR7esuKNcHSflE8ix35PjYCjHDn5Gjt1TqTH2fk01PocYewfWQTXgxGL8Wg0lfa1ejGvoC%2FEn0tDl76Ghh3In%2BFAa6g00dMUUzAry%2Be7SWWvgh9FOggagnFM70Txwgx6fXfySghrrDpIKEDB1NM2DQyVycp6%2FSSLd%2Fv5LxiXSC46LX5DTK2owqFkIAF6B7y5%2FEDx898p10UerXOy%2Bc%2BVir124zktFW7%2F2uMtl%2B8Vnjshi9OXHGNNVe%2FXVef4vgzdVO152eUc8fPpyrDAaP44SMoQHlykEOqc4dVJSl%2BOT1adgNOmHl3EfTTKsYv%2BsRTz84htHGv8qSPdP7d4f6uEmk9m9BUVU06GaCsGzjVHNTUIz4wyFzKOX4eyeY%2B85QJ0AYtLbqfGBW5J%2FspdJZwji%2F1vS%2Bb%2Bf9nwrYzRH7WvR21mcuU%2BC%2BvK74%2B85JB2%2Bnnr%2B3OuODESdB%2B2rO3KJz9HHry6ZdKcEs%2FlprwxvfiAl1%2F8B)

The pipeline pattern handles three kinds of primitives:

1. A `Generator` / `Source` that produces secret keys or seeds and does not take any inputs. In most cases, it generates either a key pair (e.g. Ed25519) or a mnemonic (BIP39) to create a seed for a deterministic wallet
2. A `Processor` that takes in locations or data, and produces some new secret, stores it in a location. It is usually used to derive a new keypair (SLIP10).
3. A `Receiver` / `Sink` that takes a location, produces a value and returns it. It does not store any product inside the vault.

All operations involving sensitive data make heavy use of the procedures framework.

### Code Example

```rust
//  .. we initialize `client` somewhere before the calls
// This constructs a `GenerateKey` procedure, that will generate a key at given
// output location in the vault
let generate_key_procedure = GenerateKey {
       ty: keytype.clone(),
       output: output_location.clone(),
};

// Even though this procedure does not create a useful output, the result can be
// used to check for errors
let procedure_result = client.execute_procedure(StrongholdProcedure::GenerateKey(generate_key_procedure));

// Front the previously generate key, we want to export the public key
let public_key_procedure = stronghold::procedures::PublicKey {
       ty: keytype,
       private_key: output_location,
};
```

````

File: documentation/docs/key_concepts/non-contiguous-data-types.md [hunk 7]
```diff
@@ -0,0 +1,48 @@
...
+<!--
+This comment shall be deleted and shall give an overview on the structure:
+
+- abstract gives an overview on memory and especially NC data types
+- memory management will be explained in general, guarded types are being specifically mentioned.
+   - libsodium is mentioned as low-level driving force
+   - the current interface for guarded memory types is presented
+- nc data types are being explained in greater detail,
+- the boojum scheme is being explored with password handling as an example  
+-->
```
lucas-tortora:
```suggestion
```

documentation/docs/key_concepts/runtime-extensions.md
```diff
@@ -0,0 +1,58 @@
...
+- apit
+- security
+- runtime
+- traits
+---
+
+# Runtime extensions 
+
+#### Authors: Matthias Kandora - \<matthias.kandora@iota.org>
+
```
lucas-tortora:
```suggestion
```

documentation/docs/key_concepts/procedures.md
```diff
@@ -0,0 +1,69 @@
...
+<!-- 
+    this note can and should be deleted. 
+    it's purpose is to highlight the structure of this document. 
+
+    - the abstracts highlights the reason for the procedures framework to access secrets
+    - in-depth 
+
+--->
+
```
lucas-tortora:
```suggestion
```

documentation/docs/key_concepts/transactional-concurrency.md
```diff
@@ -0,0 +1,26 @@
...
+- stm
+- concurrency
+- lockless
+---
+
+# Software Transactional Memory and Transactional Concurrency
+
+#### Authors: Matthias Kandora - \<matthias.kandora@iota.org>
+
+***Abstract:***
```
lucas-tortora:
```suggestion
```

documentation/docs/key_concepts/transactional-concurrency.md
```diff
@@ -0,0 +1,26 @@
+---
+description: Software Transactional Memory and Transactional Concurrency
+image: /img/logo/Stronghold_icon.png
+keywords:
+- stm
+- concurrency
+- lockless
+---
```
lucas-tortora:
```suggestion
- explanation
---
```

documentation/docs/key_concepts/p2p.md
```diff
@@ -0,0 +1,27 @@
+---
+description: Stronghold peer to peer communication 
+image: /img/logo/Stronghold_icon.png
+keywords:
+- p2p
+- networking
+---
```
lucas-tortora:
```suggestion
- explanation
---
```

documentation/docs/key_concepts/procedures.md
```diff
@@ -0,0 +1,58 @@
+---
+description: Cryptographic procedures
+image: /img/logo/Stronghold_icon.png
+keywords:
+- security
+- procedures
+- vault
+---
```
lucas-tortora:
```suggestion
- explanation
---
```

documentation/docs/key_concepts/non-contiguous-data-types.md
```diff
@@ -0,0 +1,38 @@
+---
+description: Handling secrets at runtime with non-contigouos data types. 
+image: /img/logo/Stronghold_icon.png
+keywords:
+- bojum scheme
+- non-contiguous data types
+- security
+- runtime
+---
```
lucas-tortora:
```suggestion
- explanation
---```

File: documentation/docs/key_concepts/transactional-concurrency.md [hunk 15]
```diff
@@ -0,0 +1,26 @@
...
+
+This approach guarantees that modifications on memory are always consistent, but it comes with a restriction. Since transactions can be retried, operations inside a transaction necessarily need to be idempotent and shouldn’t have any side effects. In an extreme case think of a function that launches an ICMB: you simply can’t reverse the process. Another edge case concerning STM-based approaches are interleaving transactions, where reads and writes are alternating between two threads. In a worst-case scenario both transactions would retry indefinitely.
```
lucas-tortora:
```suggestion
Writing software that runs on CPUs with more than one core is the norm nowadays. Software libraries heavily use concurrent code, separating sequential steps of work into concurrent, possibly parallel executed packages of work. Therefore, we need the means to synchronize the work on specific points to keep a consistent state across all threads. This section explains a relatively new approach for optimistic locking without the fear of deadlocks: software transactional memory.

## Software Transactional Memory

Today’s computing power derives from many CPU cores doing work in parallel. Any software that doesn’t use present-day concurrency will be less performant than software that does. Stronghold is no different, and Rust is an excellent programming language that offers a lot of concurrency and asynchronous primitives.


### The Actor Model

Stronghold employs a well known concurrency architecture: the actor model. The basic idea of the actor model is to have isolated actors, each taking care of some functionality. Actors receive messages with data to act upon and return data when they finish processing it. Since each actor contains its own state and concurrency is achieved by not directly calling functions but by polling messages, most of the undesirable concurrency problems are taken care of. Dead-locks will never occur. 

The actor system is almost ubiquitous in the target architecture. Many modern languages have the actor model built in with an excellent supervisor, among other tools. With Rust, the integration involves a lot of boilerplate code. But it’s for each user to decide whether or not the actor model approach is favorable. We wanted to know whether we could provide a simple interface, ideally some primitive types to work on with simple function calls, but still run in a concurrent setup without the headaches that come with locks and mutexes. 

## Software Transactional Memory (STM)

A newer approach is using Software Transactional Memory (STM). STMs have been around for quite some time. In STMs, each operation on memory happens in an atomic transaction. Whenever memory is modified, this modification is written into a log. While inside a transaction, reading from memory also is done through a log. The transaction has finished when all changes recorded inside the log have been committed to the actual memory. A transaction fails if another thread tries to modify the targeted piece of memory between operations. A failed transaction can be re-run any number of times. 

This approach guarantees that modifications to memory are always consistent, but it comes with a restriction. Since transactions can be retried, operations inside a transaction must be idempotent and should not have any side effects. In an extreme case, think of a function that launches an ICMB: you can not reverse the process. Another edge case concerning STM-based approaches is interleaving transactions, where reads and writes are alternating between two threads. In a worst-case scenario, both transactions would retry indefinitely. 

```

documentation/docs/key_concepts/p2p.md
```diff
@@ -0,0 +1,27 @@
...
+
+Stronghold makes use of static keys, and ephemeral keys, which will be used for the handshake protocol and the encrypted transport. The handshake pattern being used for secret exchange is being designed as XX. where the first X describes the static key transmitted to the responder, while the second describes the static key for responder transmitted by initiator. See the protocol website or in-depth explanation, on how the NOISE protocol works.
+
+Each incoming connection is secured by a simple application level firewall. Rules can be defined for each peer trying to access certain parts of a Stronghold. Eg. the firewall can be configured, that only certain paths are available to a remote requesting peer.
```
lucas-tortora:
```suggestion
# Peer to Peer Communication in Stronghold

Stronghold includes extensive communication features. At surface level, there are two modes of operations:

1. You can run Stronghold as a server to provide services on sensitive data like remote procedure execution
2. You can run Stronghold as a relay, some intermediary to connect peers. If not peer to peer connection is directly possible.

At the time of writing, all peer to peer connections are currently routed through the relay, but this restriction will be lifted in the future.

If you need some kind of server-client setup, you should use the first mode. Think of a remote security module where you can sign messages, export public keys, etc. You should use the second mode to synchronize data between two Strongholds.

The peer to peer capabilities are built on top of libp2p, the foundation of IPFS. Communication between two Strongholds is secured by the underlying implementation of the NOISE protocol. In short, in the beginning, two Strongholds exchange handshake messages. Each participant generates a key pair (e.g., Ed25519), exchanges [Diffie-Helman](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange) public keys, operates on ephemeral keys and hashes the results into a shared secret. The shared secret is later used to send encrypted transport messages.

Stronghold uses static and ephemeral keys for the handshake protocol and encrypted transport. The handshake pattern used for secret exchange is designed as XX. The first X describes the static key transmitted to the responder, while the second describes the static key for the responder sent by the initiator.

:::note Noise Protocol

You can find a detailed explanation of how the NOISE protocol works on the official [NOISE Protocol website](http://www.noiseprotocol.org/).

:::

A simple application level firewall secures each incoming connection. You can define rules for each peer trying to access certain parts of a Stronghold. For example, you can configure the firewall so only certain paths are available to a remote peer.

```

documentation/docs/key_concepts/p2p.md
```diff
@@ -0,0 +1,27 @@
...
+
+Strongholds running as remote Instances are being addressed with the multiaddr format.
```
lucas-tortora: I think there's something I'm not quite understanding in this sentence.

documentation/docs/key_concepts/runtime-extensions.md
```diff
@@ -0,0 +1,56 @@
+---
+description: Runtime extensions
+image: /img/logo/Stronghold_icon.png
+keywords:
+- apit
+- security
+- runtime
+- traits
+---
```
lucas-tortora:
```suggestion
- explanation
---
```

documentation/docs/key_concepts/p2p.md
felsweg-iota: > Strongholds running as remote Instances are being addressed with the multiaddr format.

This sentence has been updated, and should be more clear now.

lucas-tortora (commented):
@felsweg-iota , could you rename the `key_concepts` folder to `explanations` and update sidebar.js accordingly?  Other than that this looks good to me :)

documentation/docs/explanations/non-contiguous-data-types.md
```diff
@@ -0,0 +1,39 @@
...
+keywords:
+- bojum scheme
+- non-contiguous data types
+- security
+- runtime
+- explanation
+---```
+
+
+# Non-Contiguous Data Types and Handling secrets at runtime
```
lucas-tortora:
```suggestion
# Non-Contiguous Data Types and Handling Secrets at Runtime
```

documentation/docs/explanations/runtime-extensions.md
```diff
@@ -0,0 +1,57 @@
...
+image: /img/logo/Stronghold_icon.png
+keywords:
+- apit
+- security
+- runtime
+- traits
+- explanation
+---
+
+# Runtime extensions
```
lucas-tortora:
```suggestion
# Runtime Extensions 
```

documentation/sidebars.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
     items: [
       {
```
lucas-tortora, line 22:
```suggestion
      {
        type: 'doc',
        id: 'explanations/non-continuos-data-types',
        label: 'Non-continuos Data Types'
      },
      {
        type: 'doc',
        id: 'explanations/p2p',
        label: 'Peer to Peer Communication'
      },
      {
        type: 'doc',
        id: 'explanations/procedures',
        label: 'Cryptographic Procedures'
      },
      {
        type: 'doc',
        id: 'explanations/runtime-extensions',
        label: 'Runtime Extensions'
      },
      {
        type: 'doc',
        id: 'explanations/transactional-concurrency',
        label: 'Transactional Memory and Transactional Concurrency'
      },
      {
```

lucas-tortora (approved):

documentation/docs/explanations/non-contiguous-data-types.md
lucas-tortora:
```suggestion
# Non-Contiguous Data Types and Handling Secrets at Runtime
```

documentation/docs/explanations/non-contiguous-data-types.md
```diff
@@ -0,0 +1,39 @@
...
+
+Non-contiguous (NC) data types store their inner referenced data in multiple locations, either in memory, on the file system, or a mixture of both. NC data types are useful if the memory is partitioned into multiple segments, and storing a continuous stream of bytes might not be possible. The disadvantage is that the operating system must constantly keep track of the referenced memory segments.
+
+
+### Boojum Scheme
+
+Non-contiguous memory types split protected memory into multiple fragments, mitigating any memory dumps and making it virtually impossible for attackers to retrieve stored data. The following section describes non-contiguous memory types in more detail with a use case we often encountered and solved when we were developing Stronghold.
+
+#### Use Case: Passphrase Management
```
lucas-tortora:
```suggestion
#### Use case: passphrase management
```

documentation/sidebars.js
```diff
@@ -21,7 +21,32 @@ module.exports = {
...
         type: 'doc',
-        id: 'key_concepts/retrospective',
+        id: 'explanations/non-continuos-data-types',
```
lucas-tortora:
```suggestion
        id: 'explanations/non-contiguous-data-types.md',
```

documentation/sidebars.js
```diff
@@ -21,7 +21,32 @@ module.exports = {
...
+        label: 'Non-continuos Data Types'
```
lucas-tortora:
```suggestion
        label: 'Non-contiguous Data Types'
```

documentation/sidebars.js
```diff
@@ -21,7 +21,32 @@ module.exports = {
     items: [
       {
         type: 'doc',
-        id: 'key_concepts/retrospective',
+        id: 'explanations/non-contiguous-data-types.md',
```
lucas-tortora:
```suggestion
        id: 'explanations/non-contiguous-data-types',
```

lucas-tortora:
```suggestion
        id: 'explanations/non-contiguous-data-types',
```

PR #434, opened 2022-09-30
Force client to be able to be loaded only once

### LINKED ISSUE #430: Function `load_client` ambiguous behaviour
## Bug description

The function `load_client` from a `Stronghold` instance creates a new `Client` instance every time even if there is already another `Client` with the same path being used. 
This can be an issue when using Stronghold with multiple threads. Multiple threads could get different clients with the same path which can result in unpredictable behavior when commiting these clients to the snapshot.

## Rust version

Which version of Rust are you running?

- Rust version: 1.63

## Steps To reproduce the bug

Explain how the maintainer can reproduce the bug.

1. Create 2 threads
2. Call `load_client` from the two threads
3. Modify these clients
4. Commit the client of each thread
5. Changes of one thread won't be kept in the Snapshot

## Expected behaviour

Changes from both clients should be saved in the snapshot

## Actual behaviour

Only the changes of one client is left in the snapshot

vuongDang:
# Description of change
!!! This change breaks the API behaviour

- `load_client` and `load_client_from_snapshot` returns an error if client was already loaded
- add `unload_client` function to clear a client from the clients managed by a Stronghold instance 
- adapt the tests to this new behaviour
- add test for new behaviour and new function

fixes issue #430 

## Type of change

- [ ] Documentation Fix

Check for errors when clients are loaded twice, unload client and load them again successfully 

## Change checklist

### DIFF
--- a/client/src/tests/interface_tests.rs
+++ b/client/src/tests/interface_tests.rs
@@ -174,9 +174,8 @@ fn test_stronghold_purge_client() {
 
     assert!(client.record_exists(&output_location).unwrap());
     assert!(client2.record_exists(&output_location).unwrap());
+    assert!(stronghold.unload_client(client2).is_ok());
 
-    // Reload client2 from the snapshot state.
-    std::mem::drop(client2);
     let client2 = stronghold.load_client(&client_path2).unwrap();
 
     stronghold.purge_client(client).unwrap();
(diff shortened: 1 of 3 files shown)


felsweg-iota (approved):
good idea to explicitly check for an already loaded client.

PR #440, opened 2022-10-05
Stm documentation and cleanup of code

vuongDang:
# Description of change

- Cleanup on the work of concurrent Stronghold to be reusable in the future
- Documention on concurrency in the wiki

## Type of change

-  Documentation Fix

## Change checklist


stm/README.md
```diff
@@ -0,0 +1,54 @@
+# STM 
+
+This crate contains different implementation to have a concurrency in Stronghold.
+Two approaches are implemented: Read-Log-Update (RLU) and Transactional Locking 2 (TL2).
```
lucas-tortora:
```suggestion
This crate contains two different implementations to have a concurrency in Stronghold.
We have implemented two approaches:

* [Transactional Locking 2 (TL2)](#software-transactional-memory-stm-with-tl2).
* [Read-Log-Update (RLU)](#read-log-update) 
```

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+~In an STM based system all objects having mutable state are transactional, the behavior on the objects are transparently using the underlying system.
+This allows to isolate guarded memory from being exposed at runtime.
+Transactions are always safe; internal conflicts are rolled back automatically and retried until the transaction succeeds.
+Operations on mutable memory are composable.
+Since only write operations actually change the object, this operation must be communicated to the other threads.
+Recent work describes multiple approaches, where we consider blocking / retrying other transaction the most viable approach to ensure data consistency.
+If the transaction has been finished, while having all read operations validated, the resulting work is committed to the actual object.
+Other threads operating on the same object, never see the changes done to this object.
+The STM described here, uses a lazy approach in rolling back a transaction An STM can also be described as an optimistic locking approach: Work on memory is considered safe until a conflict occurs, the transaction can be safely rolled back and retried.~
```
lucas-tortora:
```suggestion
~In an STM-based system, where all objects with a mutable state are transactional, the behavior of the objects is transparent using the underlying system. This isolates guarded memory, keeping it from being exposed at runtime. Transactions are always safe; internal conflicts are rolled back automatically and retried until the transaction succeeds. Operations on mutable memory are composable. Since only write operations can change the object, these operations must be communicated to the other threads.

Recent work describes multiple approaches, where we consider blocking or retrying other transactions. This is the most viable approach to ensure data consistency. The resulting work is committed to the actual object if the transaction is finished while having all read operations validated; other threads operating on the same object never see the changes done to this object.

The STM described here uses a lazy approach in rolling back a transaction. An STM can also be described as an optimistic locking approach: work on memory is considered safe until a conflict occurs, and the transaction can be safely rolled back and retried.~
```

I assume it's intentional, but just in case, this has a strike-through.

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+We chose to implement the TL2 implementation of STM which implementation details can be found in [[1]](https://citeseer.ist.psu.edu/viewdoc/summary?doi=10.1.1.90.811&rank=4&q=various%20cross%20version%20operation&osm=&ossid=).
+Improvements and tests that have been applied to TL2 are presented in [[2]](https://www.researchgate.net/publication/220854689_Testing_patterns_for_software_transactional_memory_engines).
```
lucas-tortora:
```suggestion
We chose to implement the TL2 implementation of STM. You can find the implementation details in the [Transactional Locking II (2006) paper](https://citeseer.ist.psu.edu/viewdoc/summary?doi=10.1.1.90.811&rank=4&q=various%20cross%20version%20operation&osm=&ossid=).
The improvements and tests that have been applied to TL2 are presented in [Testing patterns for software transactional memory engines paper](https://www.researchgate.net/publication/220854689_Testing_patterns_for_software_transactional_memory_engines).
```

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### Current flaws 
+- Variables that can cloned and shared between thread are classified into the enum `SharedValue` 
+  - this is a bit ugly since we need to import types of Stronghold into this crate to use our STM implementation
+  - we tried to do a generic implementation using trait object (shared types needs to implement a trait to be used in STM)
+    but we were not successful. This was mostly due to the fact that shared variable needed to implement `Clone: Sized` for STM
+    but since they are trait objects they also need to implement `!Sized`. Implementation of both is incompatible.
+- This is slow, tests on Stronghold show most of the time going single-treaded was faster than multi-threaded (x10)
```
lucas-tortora:
```suggestion

- Variables that can be cloned and shared between threads are classified into the enum `SharedValue.`
  - This is not optimal since we need to import types of Stronghold into this crate to use our STM implementation.
  - We tried to do a generic implementation using trait object (shared types need to implement a trait to be used in STM)
    but we were not successful. This was mostly because the shared variable needed to implement `Clone: Sized` for STM
    , but since they are trait objects, they also need to implement `!Sized`. Implementing both is incompatible.
- This is slow. Tests on Stronghold show most of the time going single-threaded was faster than multi-threaded (x10).
```

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Read-Log-Update
+
+RLU is an approach similar to STM with the main difference that transactions are never rolled back but blocked until they can be committed to produce a consistent state.
+Implementation details of RLU can be found in [[1]](https://sigops.org/s/conferences/sosp/2015/current/2015-Monterey/printable/077-matveev.pdf).
```
lucas-tortora:
```suggestion
RLU is an approach similar to STM, with the main difference being that transactions are never rolled back but blocked until they can be committed, producing a consistent state.
You can find Implementation details of RLU in the [Read-Log-Update paper](https://sigops.org/s/conferences/sosp/2015/current/2015-Monterey/printable/077-matveev.pdf).
```

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### Current flaws 
+- Data race and deadlocks occur in the tests
```
lucas-tortora:
```suggestion

- Data races and deadlocks occur in the tests.
```

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Features
+
+[x] - multiple concurrent reads and writes
+[x] - lock free integration
+[x] - (optional) protected memory handling integrated
```
lucas-tortora:
```suggestion
[x] - Multiple concurrent reads and writes.
[x] - Lock free integration.
[x] - (optional) Protected memory handling integrated.
```

stm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+# Open Issues / Todo
+
+- Feature gated secured memory 
+- RLU is global context and must be accessible as mutable from many threads, as new contexts can be spawned any time
+- check if try_lock needs to return a full copy for the log, or if the change has to be delayed
```
lucas-tortora:
```suggestion
- Feature-gated secured memory.
- RLU is a global context and must be accessible as mutable from many threads, as new contexts can be spawned anytime.
- Check if `try_lock` needs to return a full copy for the log or if the change has to be delayed.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
+---
+description: Stronghold as a concurrent software
+image: /img/logo/Stronghold_icon.png
+keywords:
+- stm
+- concurrency
+- lockless
+- explanation
```
lucas-tortora:
```suggestion
description: Learn about Stronghold as a concurrent software, the actor model STM with TL2 and RLU. 
image: /img/logo/Stronghold_icon.png
keywords:
- actor model
- rlu
- STM
- concurrency
- lockless
- explanation
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+---
+
+# Concurrency in Stronghold
+
+Writing software that runs on CPUs with more than one core is the norm nowadays. Software libraries heavily use concurrent code, separating sequential steps of work into concurrent, possibly parallel executed packages of work.
+Writing concurrent programs makes the most of our current computers and servers in terms of performance. The downsides of such approach is that it is difficult to write such programs without having bugs. This is mostly due to a problem called _data race_ where multiple threads temper the same shared memory at the same time. This may create inconsistent behaviour of the concurrent program due to the fact that each thread executes its computations in a non-deterministic order.
+Multiple solutions to tackle concurrent programming exist and we have been exploring them for Stronghold.
+In the following sections we will present you different solutions that have been tested to have Stronghold as a concurrent library.
```
lucas-tortora:
```suggestion
Writing software that runs on CPUs with more than one core is the norm nowadays. Software libraries heavily use concurrent code, separating sequential steps of work into concurrent, possibly parallel, work packages.

Writing concurrent programs makes the most of our current computers and servers in terms of performance. The downside of this approach is that it is difficult to write such programs without having bugs. This is mostly due to a problem called _data race_, where multiple threads simultaneously tamper with the same shared memory, which may create inconsistent behavior of the concurrent program because each thread executes its computations in a non-deterministic order.

Multiple solutions to tackle concurrent programming exist, and we have been exploring them for Stronghold.
In the following sections, we will present different solutions tested to make Stronghold a concurrent library.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+## The Actor Model
+
+Stronghold employs a well known concurrency architecture: the actor model. The basic idea of the actor model is to have isolated actors, each taking care of some functionality. Actors receive messages with data to act upon and return data when they finish processing it. Since each actor contains its own state and concurrency is achieved by not directly calling functions but by polling messages, most of the undesirable concurrency problems are taken care of. Dead-locks will never occur.
```
lucas-tortora:
```suggestion
Stronghold employed a well-known concurrency architecture: the actor model. The basic idea of the actor model is to have isolated actors, each taking care of some functionality. Actors receive messages with data to act upon and return data when they finish processing it. Since each actor contains its own state and concurrency is achieved by not directly calling functions, but by polling messages, most undesirable concurrency problems are taken care of. Deadlocks will never occur.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+In the following sections we will present you different solutions that have been tested to have Stronghold as a concurrent library.
+
+## The Actor Model
+
+Stronghold employs a well known concurrency architecture: the actor model. The basic idea of the actor model is to have isolated actors, each taking care of some functionality. Actors receive messages with data to act upon and return data when they finish processing it. Since each actor contains its own state and concurrency is achieved by not directly calling functions but by polling messages, most of the undesirable concurrency problems are taken care of. Dead-locks will never occur. 
+
+The actor system is almost ubiquitous in the target architecture. Many modern languages have the actor model built in with an excellent supervisor, among other tools. With Rust, the integration involves a lot of boilerplate code. But it’s for each user to decide whether or not the actor model approach is favorable. We wanted to know whether we could provide a simple interface, ideally some primitive types to work on with simple function calls, but still run in a concurrent setup without the headaches that come with locks and mutexes.
```
lucas-tortora:
```suggestion
The actor system is almost ubiquitous in the target architecture. Many modern languages have a built-in actor model with an excellent supervisor, among other tools. With Rust, the integration involves a lot of boilerplate code. But it’s for each user to decide whether or not the actor model approach is favorable. We wanted to know whether we could provide a simple interface, ideally some primitive types to work on with simple function calls, but still run in a concurrent setup without the headaches that come with locks and mutexes.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Why didn't we keep the actor model?
+
+First iteration of Stronghold with the actor model was done with the rust crate called [Riker](https://riker.rs/) actor system crate. 
+Unfortunately this crate was discontinued we had to find an alternative which was the [Actix](https://github.com/actix/actix) actor framework crate.
+Actix uses the Tokio runtime which came as an issue for us as the Stronghold library. 
+Indeed software using Stronghold had to pass ownership of the runtime to our code when making function calls to the Stronghold library.
+This was cumbersome for our users and we abandoned the idea of using the actor model for concurrent programming and explored other paradigms.
```
lucas-tortora:
```suggestion
The first iteration of Stronghold implemented the actor model with the rust crate [Riker](https://riker.rs/), an actor system crate.

Unfortunately, since this crate was discontinued, we had to find an alternative: the [Actix](https://github.com/actix/actix) actor framework crate.
Actix uses the Tokio runtime, which came as an issue for us as the Stronghold library.

Software using Stronghold had to pass ownership of the runtime to our code when making function calls to the Stronghold library.
This was cumbersome for our users, and we abandoned the idea of using the actor model for concurrent programming and explored other paradigms.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+## Software Transactional Memory (STM) with Transactional Locking 2 (TL2)
+
+A newer approach is using Software Transactional Memory (STM). STMs have been around for quite some time. In STMs, each operation on memory happens in an atomic transaction. Whenever memory is modified, this modification is written into a log. While inside a transaction, reading from memory also is done through a log. The transaction has finished when all changes recorded inside the log have been committed to the actual memory. A transaction fails if another thread tries to modify the targeted piece of memory between operations. A failed transaction can be re-run any number of times.
```
lucas-tortora:
```suggestion
STMs have been around for quite some time. In STMs, each operation on memory happens in an atomic transaction. Whenever memory is modified, this modification is written into a log. While inside a transaction, reading from memory is also done through a log. The transaction has finished when all changes recorded inside the log have been committed to the actual memory. A transaction fails if another thread tries to modify the targeted piece of memory between operations. A failed transaction can be re-run any number of times.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+A newer approach is using Software Transactional Memory (STM). STMs have been around for quite some time. In STMs, each operation on memory happens in an atomic transaction. Whenever memory is modified, this modification is written into a log. While inside a transaction, reading from memory also is done through a log. The transaction has finished when all changes recorded inside the log have been committed to the actual memory. A transaction fails if another thread tries to modify the targeted piece of memory between operations. A failed transaction can be re-run any number of times. 
+
+This approach guarantees that modifications to memory are always consistent, but it comes with a restriction. Since transactions can be retried, operations inside a transaction must be idempotent and should not have any side effects. In an extreme case, think of a function that launches an ICMB: you can not reverse the process. Another edge case concerning STM-based approaches is interleaving transactions, where reads and writes are alternating between two threads. In a worst-case scenario, both transactions would retry indefinitely.
```
lucas-tortora:
```suggestion
This approach guarantees that modifications to memory are always consistent, but it comes with a restriction. Since transactions can be retried, operations inside a transaction must be idempotent and should not have any side effects. In an extreme case, think of a function that launches an ICMB: you can not reverse the process. Another edge case concerning STM-based approaches is interleaving transactions, where reads and writes are alternating between two threads. In a worst-case scenario, both transactions would retry indefinitely.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+This approach guarantees that modifications to memory are always consistent, but it comes with a restriction. Since transactions can be retried, operations inside a transaction must be idempotent and should not have any side effects. In an extreme case, think of a function that launches an ICMB: you can not reverse the process. Another edge case concerning STM-based approaches is interleaving transactions, where reads and writes are alternating between two threads. In a worst-case scenario, both transactions would retry indefinitely. 
+
+### TL2
+
+Implementation of TL2 is inspired from these papers [1](https://citeseer.ist.psu.edu/viewdoc/summary?doi=10.1.1.90.811&rank=4&q=various%20cross%20version%20operation&osm=&ossid=) and [2](https://www.researchgate.net/publication/220854689_Testing_patterns_for_software_transactional_memory_engines).
```
lucas-tortora:
```suggestion

The Implementation of TL2 is inspired by these papers:

* [Transactional Locking II (2006)](https://citeseer.ist.psu.edu/viewdoc/summary?doi=10.1.1.90.811&rank=4&q=various%20cross%20version%20operation&osm=&ossid=)
* [Testing patterns for software transactional memory engines](https://www.researchgate.net/publication/220854689_Testing_patterns_for_software_transactional_memory_engines).
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+The idea is quite straightforward. A global clock is used and incremented every time a transaction terminates. 
+Shared memory/variables are tagged with a local clock value corresponding to the moment they were last modified by a transaction commitment.
+Transactions are computed on copies of these shared memory/variables.
+At the end of each transaction the system needs to check that all the shared memory/variables used during the transaction have not been tampered by another transaction to keep a consistent behaviour. 
+To check such thing the local clock of the copies used for the transaction are checked against the local clock of the original shared memory/variables.
+If the check detect any inconsistency between clock values then the transaction is reset and computed again.
+If the check is successful the copies are committed and replace the original shared memory/variables and they are also given a new local clock value by the global clock.
```
lucas-tortora:
```suggestion
The idea is quite straightforward. A global clock is used and incremented every time a transaction terminates.
Shared memory/variables are tagged with a local clock value corresponding to the moment they were last modified by a transaction commitment.
Transactions are computed on copies of these shared memory/variables.

At the end of each transaction, the system needs to check that all the shared memory/variables used during the transaction have not been tampered with by another transaction to keep a consistent behavior.
Stronghold verifies this by checking the local clock of the copies used for the transaction against the local clock of the original shared memory/variables.

If the check detects any inconsistency between clock values, then the transaction is reset and computed again.

If the check is successful, the copies are committed and replace the original shared memory/variables, and the global clock also gives them a new local clock value.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+This approach advantage is that the idea is relatively simple and implement the STM model. The downside is that the implementation we developed of TL2 is painfully slow and usually going single thread is faster than using multiple threads. 
+We suspect that this poor performance are due to multiple interleaving of threads which force the transactions to be reset multiple times. Further investigation is required to confirm this idea.
```
lucas-tortora:
```suggestion
This approach's advantage is that the idea is relatively simple and implements the STM model. The downside is that the implementation we developed of TL2 is painfully slow, and usually going single thread is faster than using multiple threads.
We suspect this poor performance is due to multiple interleaving of threads which force the transactions to be reset multiple times. Further investigation is required to confirm this idea.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+Contrary to TL2 previously presented, RLU is a blocking algorithm.
+Threads do their computations locally and synchronize with each other using their local clock (when the computations started) and a common global clock.
+A thread logs the new value of any shared memory that it has modified.
+To read shared memory a thread either does it directly in the memory or go fetch it in another thread logs when it has been locked and is being modified.
+When a thread wants to commit its computations to the memory it has to wait that all the other threads that have started with an old memory state to terminate beforehand.
+The upside of such approach is that a thread never needs to redo its computation since it has waited the correct timing to commit its work.
+In return threads have to spend some time idle since they have to wait. 
+However the whole RLU algorithm is lock-free meaning that at any moment there is always a thread that advances in its computation and the whole system cannot be hard stuck.
```
lucas-tortora:
```suggestion
Contrary to [TL2](#tl2), RLU is a blocking algorithm.
Threads do their computations locally and synchronize with each other using their local clock (when the computations started) and a common global clock.
A thread logs the new value of any shared memory it has modified.

To read shared memory, a thread either does it directly in the memory or fetches it in another thread logs when it has been locked and is being modified.
When a thread wants to commit its computations to the memory, it has to wait for all the other threads that have started with an old memory state to terminate.

The upside to this approach is that a thread never needs to redo its computation since it has to wait for the correct timing to commit its work.

In return, threads may be idle for some time as they have to wait.
However, the whole RLU algorithm is lock-free. This means that at any moment there is always a thread that advances in its computation, and the whole system cannot be hard-stuck.
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Concurrent programming with locks
```
lucas-tortora:
```suggestion
## Concurrent Programming with Locks
```

documentation/docs/explanations/concurrency.md
```diff
@@ -0,0 +1,80 @@
...
+
+Using locks is one of the most basic approach to concurrent programming. 
+Lock are used to control the access to shared memory/variables. Depending of the type of locks used you can restrict the access to a shared memory/variable to either one or a limited number of threads. 
+This prevents data race since the different threads cannot access the shared memory/variables in a chaotic way. 
+The biggest issue is that those locks are in general difficult to use correctly and really hard to debug. 
+One of the most problematic situation that can happen with locks are __deadlocks__. 
+Deadlocks happen when the whole system cannot advance anymore due to the fact that different threads require some locks to advance their computation but these locks are kept and blocked by other threads in a similar situation.
+Models like the actor system or the STM have been created to use locks as little as possible.
+
+In our case the API of Stronghold is relatively straightforward therefore we're exploring the idea of using basic locks instead of a resource heavy concurrent framework.
```
lucas-tortora:
```suggestion
Using locks is one of the most basic approaches to concurrent programming.
Locks are used for controlling access to shared memory/variables. Depending on the type of locks used, you can restrict access to a shared memory/variable to either one or a limited number of threads.
This prevents data race since the different threads cannot access the shared memory/variables chaotically.
The biggest issue is that those locks are generally difficult to use correctly and hard to debug.

One of the most problematic situations with locks is a __deadlock__.
Deadlocks happen when the whole system cannot advance anymore because different threads require some locks to advance their computation, but these locks are kept and blocked by other threads in a similar situation.
Models like the [actor system](the-actor-model) or the [STM](#software-transactional-memory-stm-with-transactional-locking-2-tl2) use locks as little as possible.

In our case, Stronghold’s API is relatively straightforward, so we're exploring the idea of using basic locks instead of a resource-heavy concurrent framework.
```

lucas-tortora (commented):
A few suggestions regarding structure and style.  Awesome contribution :) Thanks!

vuongDang:
> A few suggestions regarding structure and style. Awesome contribution :) Thanks!

Wow you even read the README.md, thanks for the suggestions

stm/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+pub mod rlu;
+pub mod stm;
+
+// public re-exports
+// pub use breaker::BusyBreaker;
```
felsweg-iota, line 8: unused imports can be removed.

stm/script.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
felsweg-iota: do we need this bash script, or could we do this inside the tests?

stm/src/lib.rs
felsweg-iota: unused imports can be removed.

stm/src/rlu/guard/write.rs
```diff
@@ -1,6 +1,6 @@
-// Copyright 2020-2022 IOTA Stiftung
+// Copyright 2020-2021 IOTA Stiftung
```
felsweg-iota:
```suggestion
// Copyright 2020-2022 IOTA Stiftung
```

felsweg-iota (approved):

PR #474, opened 2023-05-19
Snapshot migration v3age 2.0

semenov-vladyslav:
# Description of change

This PR fixes a number of issues in stronghold including insecure snapshot encryption format is replaced with password-based [age-encryption.org/v1](https://age-encryption.org/v1) format, secrets leaked into stack/heap memory are now zeroized after use.

Be sure to reference any related issues by adding `fixes issue #`.

## Type of change

Tests now take longer due to secure password-based key derivation which should take approx 1 second on a modern computer. To alleviate it in tests set work_factor parameter to 1 in snapshot encryption routines.

## Change checklist


thibault-martinez (changes requested):
Clippy and format

client/src/types/snapshot.rs
```diff
@@ -127,15 +127,18 @@ impl Display for SnapshotPath {
...
 #[derive(Clone, Debug)]
 pub enum UseKey {
-    Key(snapshot::Key),
+    Key(Zeroizing<snapshot::Key>),
     Stored(Location),
 }
 
 impl Snapshot {
+    /// Age work factor used with strong passwords (random keys).
+    const STRONG_KEY_WORK_FACTOR: u8 = 0;
```
Thoralf-M, line 136: Shouldn't this be https://github.com/iotaledger/crypto.rs/blob/2c455b182d02ba7b354487b2fd6e9b7443fbd179/src/keys/age.rs#L640?

client/src/types/snapshot.rs
```diff
@@ -212,24 +215,24 @@ impl Snapshot {
...
-                data.try_into().map_err(|_| SnapshotError::SnapshotKey(vid, rid))?
+                Ok(self
+                    .db
+                    .get_guard::<_, SnapshotError, _>(&pkey, vid, rid, |guarded_data| {
+                        let key_ref = guarded_data.borrow();
+                        let key: &snapshot::Key = (*key_ref)
+                            .try_into()
+                            .map_err(|_| SnapshotError::SnapshotKey(vid, rid))?;
+                        Ok(snapshot::encrypt_file(&data, snapshot_path.as_path(), key, &[])?)
+                    })?)
```
Thoralf-M: Can the `Ok(..?)` not be removed?
```suggestion
                self
                    .db
                    .get_guard::<_, SnapshotError, _>(&pkey, vid, rid, |guarded_data| {
                        let key_ref = guarded_data.borrow();
                        let key: &snapshot::Key = (*key_ref)
                            .try_into()
                            .map_err(|_| SnapshotError::SnapshotKey(vid, rid))?;
                        snapshot::encrypt_file(&data, snapshot_path.as_path(), key, &[])
                    })
```

client/src/types/snapshot.rs
semenov-vladyslav: No, strong password means 128-bit entropy key which can be used for encryption straightaway without any strengthening. In practice it doesn't degrade security but improves performance. RECOMMENDED_MINIMUM_ENCRYPT_WORK_FACTOR is for passwords.

engine/src/snapshot/logic.rs
```diff
@@ -59,120 +63,120 @@ pub enum WriteError {
...
 /// This is achieved by creating a temporary file in the same directory as the specified path (same
 /// filename with a salted suffix). This is currently known to be problematic if the path is a
 /// symlink and/or if the target path resides in a directory without user write permission.
-pub fn write_to(plain: &[u8], path: &Path, key: &Key, associated_data: &[u8]) -> Result<(), WriteError> {
+pub fn encrypt_file(plain: &[u8], path: &Path, key: &Key, associated_data: &[u8]) -> Result<(), WriteError> {
     // TODO: if path exists and is a symlink, resolve it and then append the salt
     // TODO: if the sibling tempfile isn't writeable (e.g. directory permissions), write to
+    if !associated_data.is_empty() {
+        return Err(WriteError::UnsupportedAssociatedData);
+    }
```
Thoralf-M: Why is associated_data even a parameter if it's not supported?

engine/src/snapshot/migration/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+        Self::V2 { path, key, aad }
+    }
+
+    pub fn v3(path: &'a Path, password: &'a [u8]) -> Self {
+        Self::V3 { path, password }
+    }
+}
+
+/// Magic bytes (bytes 0-4 in a snapshot file) aka PARTI
+const MAGIC: [u8; 5] = [0x50, 0x41, 0x52, 0x54, 0x49];
```
Thoralf-M, line 49: Any specific reason why it's exactly these and not something else?

Thoralf-M: Nvm is there for historic reasons

semenov-vladyslav: I have no clue. That's the old magic bytes.

engine/src/snapshot/logic.rs
semenov-vladyslav: Good point. Initially I didn't want to break the API. Now we can get rid of it.

client/src/types/snapshot.rs
semenov-vladyslav: This trick was used to easily convert between error types, rust does it automatically with From trait. Not sure if it's required here. Otherwise we'd have to use map_err explicitly.

Thoralf-M: If it returns different errors then it's fine to do this for the conversion :+1:

semenov-vladyslav: No, it's required here.

engine/src/snapshot/logic.rs
semenov-vladyslav: Done. Removed it.

Thoralf-M: Nice, but should also get a changelog entry

client/src/security/keyprovider.rs
```diff
@@ -77,25 +77,18 @@ impl KeyProvider {
     ///
     /// assert_eq!(key, expected, "Key does not match expected {:?}", key);
     /// ```
+    #[deprecated = "This constructor promotes misuse of long passphrases"]
```
PhilippGackstatter:
```suggestion
    #[deprecated = "this constructor promotes misuse of long passphrases"]
```

Nit: Error messages are lowercase by convention and this probably also applies here.

Does it actually make sense to deprecate something in a major version change? If someone is using it right now, they can also migrate away from it, so it shouldn't be a problem, right?

client/src/security/keyprovider.rs
```diff
@@ -258,6 +244,7 @@ impl KeyProvider {
     ///
     /// assert_eq!(key, expected);
     /// ```
+    #[deprecated = "This constructor promotes misuse of salt in KDF"]
```
PhilippGackstatter:
```suggestion
    #[deprecated = "this constructor promotes misuse of salt in KDF"]
```

client/src/types/error.rs
```diff
@@ -226,6 +226,12 @@ impl From<EngineReadError> for SnapshotError {
                 "Unsupported version: expected {:?}, found {:?}.",
                 expected, found
             )),
+            EngineReadError::UnsupportedAssociatedData => {
+                SnapshotError::Engine("Unsupported snapshot associated data".into())
```
PhilippGackstatter:
```suggestion
                SnapshotError::Engine("unsupported snapshot associated data".into())
```

client/src/types/error.rs
```diff
@@ -236,6 +242,10 @@ impl From<EngineWriteError> for SnapshotError {
             EngineWriteError::Io(io) => SnapshotError::Io(io),
             EngineWriteError::CorruptedData(e) => SnapshotError::CorruptedContent(e),
             EngineWriteError::GenerateRandom(_) => SnapshotError::Io(std::io::ErrorKind::Other.into()),
+            EngineWriteError::UnsupportedAssociatedData => {
+                SnapshotError::Engine("Unsupported snapshot associated data".into())
```
PhilippGackstatter:
```suggestion
                SnapshotError::Engine("unsupported snapshot associated data".into())
```

client/src/types/error.rs
```diff
@@ -236,6 +242,10 @@ impl From<EngineWriteError> for SnapshotError {
...
+            }
+            EngineWriteError::IncorrectWorkFactor => SnapshotError::Engine("Incorrect work factor".into()),
```
PhilippGackstatter:
```suggestion
            EngineWriteError::IncorrectWorkFactor => SnapshotError::Engine("incorrect work factor".into()),
```

client/src/types/snapshot.rs
```diff
@@ -338,16 +340,20 @@ impl Snapshot {
             .get_key(vid)
             .ok_or_else(|| SnapshotError::Inner("Missing local secret key.".to_string()))?;
```
PhilippGackstatter:
```suggestion
            .ok_or_else(|| SnapshotError::Inner("missing local secret key.".to_string()))?;
```

engine/runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -125,18 +125,17 @@ impl NonContiguousMemory {
...
+        xor_mut(&mut r, &d, NC_DATA_SIZE);
+        let hd = &blake2b::Blake2b256::digest(&r);
 
-        let new_data2 = xor(&old_data2, hash_of_old_shard1, NC_DATA_SIZE);
-        let new_data2 = xor(&new_data2, hash_of_new_shard1, NC_DATA_SIZE);
+        xor_mut(&mut m, hd, NC_DATA_SIZE);
+        xor_mut(&mut m, hr, NC_DATA_SIZE);
 
-        let (shard1, shard2) = MemoryShard::new_shards(&new_data1, &new_data2, &self.config)?;
+        let (shard1, shard2) = MemoryShard::new_shards(&r, &m, &self.config)?;
```
PhilippGackstatter: For future readability, can you give the short variable names more descriptive names?

engine/src/snapshot/logic.rs
```diff
@@ -59,120 +63,105 @@ pub enum WriteError {
...
+///
+/// This function expects key to be password-based (blake2b hash of the user provided password).
+/// It uses recommended work factor (approx. 20) to derive encryption key.
+/// It is safe to use with strong keys, although computing resources may be wasted.
+/// In this case it is recommended to use `encrypt_content_with_work_factor` with small/zero work factor.
+pub fn encrypt_content<O: Write>(plain: &[u8], output: &mut O, key: &Key) -> Result<(), WriteError> {
+    let work_factor = age::RECOMMENDED_MINIMUM_ENCRYPT_WORK_FACTOR;
+    // TODO: work_factor is intentionally 0 just for development.
+    // Use proper value in production.
+    // let work_factor = 1;
```
PhilippGackstatter: Leftover TODO?

engine/src/snapshot/migration/error.rs
```diff
@@ -0,0 +1,57 @@
...
+    RngFailed,
+    /// Age format error.
+    #[error("Age format error")]
+    AgeFormatError(age::DecError),
+    /// I/O error.
+    #[error("I/O error")]
+    IoError(std::io::Error),
+    /// Crypto error.
+    #[error("Crypto error")]
+    CryptoError(crypto::Error),
```
PhilippGackstatter: Nit: Ideally these error messages should be lowercase.

engine/src/snapshot/migration/v3.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut age = Vec::new();
+    input.read_to_end(&mut age)?;
+    age::decrypt_vec(password, age::RECOMMENDED_MAXIMUM_DECRYPT_WORK_FACTOR, &age[..]).map_err(From::from)
+}
+
+pub(crate) fn write<O: Write>(
+    plain: &[u8],
+    output: &mut O,
+    password: &Key,
+    _associated_data: &[u8],
```
PhilippGackstatter: Deliberately ignored variable? If so, can you add a comment why?

PhilippGackstatter (approved):
Looks good to me.

engine/src/snapshot/logic.rs
semenov-vladyslav: Added changelog entry.

client/src/security/keyprovider.rs
semenov-vladyslav: Thanks, didn't know about lower case in error messages.
I was reluctant about removing these weird constructors from KeyProvider because I don't understand what else they are used for. I would remove a lot of stuff from stronghold. In this particular case it's possible to use it in a safe manner, but it's not obvious how. Migration only deals with snapshot file format, but KeyProvider might be used for other things, such as providing keys as the name suggests.

client/src/security/keyprovider.rs
semenov-vladyslav: Done.

client/src/types/error.rs
semenov-vladyslav: Do you suggest changing only some specific messages to lower case or all of them? Error messages in stronghold were originally in capital case. Not sure if this is a place to change that. Also, @thibault-martinez said that they use error messages to handle some error cases, changing error message case might break correct matching of errors.

engine/runtime/src/memories/noncontiguous_memory.rs
semenov-vladyslav: I don't think it makes sense to do it in this place. It is a simple cryptographic algorithm, variables are not usually meaningful there. I can rename `m` to `message_xored_with_hash_of_randomness`, or `hr` to `hash_of_new_randomness`, would it be more readable? Other names might be misleading. Please, do suggest your options.

engine/src/snapshot/logic.rs
semenov-vladyslav: Tests (in client) now work slowly because of recommended work factor in the encryption function (in engine). I use this comment to change work factor so that tests work faster. Is it possible to do that automatically? `#[cfg(test)]` doesn't work because it is a different crate (tests run in client, but the slow code is in engine).

I can simply remove the "TODO:" part of the comment.

engine/src/snapshot/migration/error.rs
semenov-vladyslav: I think it should be safe to do it here in migration. @thibault-martinez , what do you think?

engine/src/snapshot/migration/v3.rs
semenov-vladyslav: No, just forgot to remove it. My bad. Fixed it.

client/src/types/error.rs
PhilippGackstatter: > The error message given by the Display representation of an error type should be lowercase without trailing punctuation, and typically concise.

I would suggest changing all of them to follow this [wider Rust convention](https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=lowerca#error-types-are-meaningful-and-well-behaved-c-good-err), if possible in a non-breaking manner, otherwise not breaking things takes priority.

engine/runtime/src/memories/noncontiguous_memory.rs
PhilippGackstatter: Personally, I would prefer the longer names you mentioned like `hash_of_new_randomness` and such. As I don't know the details of this algorithm I can't really suggest better names without diving into the details.

I think longer, meaningful variables always make code easier to read, but I'll leave this to your judgment.

engine/src/snapshot/logic.rs
PhilippGackstatter: I don't think it's possible to do what you want to do without making this public and allow other crates to overwrite it.

Yeah, maybe just remove the `TODO` then.

client/src/security/keyprovider.rs
PhilippGackstatter: Fair argument. I don't have a strong opinion on this. Deprecating the function might be a decent intermediate solution to deter usage but still allow existing users to continue using it.

client/src/types/error.rs
semenov-vladyslav: Sure, I need a confirmation from @thibault-martinez

semenov-vladyslav: Is it possible to configure clippy to catch these?

engine/src/snapshot/logic.rs
semenov-vladyslav: Removed.

engine/runtime/src/memories/noncontiguous_memory.rs
semenov-vladyslav: I'd keep short names. It's just 8 lines of pretty simple code.

client/src/types/error.rs
semenov-vladyslav: Done

engine/src/snapshot/migration/error.rs
semenov-vladyslav: Thibault is ok with it. Done.

client/src/types/error.rs
PhilippGackstatter: I'm not aware that this is possible. It seems very hard to do if you use thiserror macros to implement `Display` for your error, for instance.

engine/runtime/src/memories/noncontiguous_memory.rs
PhilippGackstatter: Sure, feel free to keep them as well!

client/src/types/error.rs
PhilippGackstatter: I'm not aware that this is possible. It seems very hard to do if you use thiserror macros to implement `Display` for your error, for instance.

client/examples/repl/command.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2020-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+#![allow(deprecated)]
```
DaughterOfMars: What's this for?

engine/runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -90,14 +90,14 @@ impl LockedMemory for NonContiguousMemory {
     /// Unlocks the memory and returns an unlocked Buffer
     /// To retrieve secret value you xor the hash contained in shard1 with value in shard2
     fn unlock(&self) -> Result<Buffer<u8>, MemoryError> {
-        let (data1, data2) = self.get_shards_data()?;
-        let data1 = &blake2b::Blake2b256::digest(data1);
-        let reconstructed_data = xor(data1, &data2, NC_DATA_SIZE);
+        let (r, mut m) = self.get_shards_data()?;
```
DaughterOfMars: more descriptive variables?

engine/src/vault/base64.rs
```diff
@@ -41,7 +41,7 @@ impl Base64 {
             Err(e) => {
                 let error = e.utf8_error();
                 let valid_up_to = error.valid_up_to();
-                let error_msg = format!("Fail encoding to base64: valid_up_to({})", valid_up_to);
+                let error_msg = format!("fail encoding to base64: valid_up_to({})", valid_up_to);
```
DaughterOfMars: failure

client/src/procedures/clientrunner.rs
```diff
@@ -52,22 +54,10 @@ impl Runner for Client {
     where
         F: FnOnce([Buffer<u8>; N]) -> Result<T, FatalProcedureError>,
     {
-        let mut ret = None;
-        let execute_procedure = |guard: [Buffer<u8>; N]| {
-            ret = Some(f(guard)?);
-            Ok(())
-        };
```
thibault-martinez, line 59: What's happening here? How come this can be removed?

client/examples/repl/command.rs
semenov-vladyslav: To silent clippy. it uses deprecated function.

DaughterOfMars: Any reason to not simply delete it?

engine/runtime/src/memories/noncontiguous_memory.rs
semenov-vladyslav: We've discussed this. See https://github.com/iotaledger/stronghold.rs/pull/474#discussion_r1201537073. Do you have suggestions?

client/src/procedures/clientrunner.rs
semenov-vladyslav: Callback function `f` now returns value of type `T`, `db.get_guards` uses it to return value.

engine/src/vault/base64.rs
semenov-vladyslav: Fixed.

engine/runtime/src/memories/noncontiguous_memory.rs
DaughterOfMars: No, because I don't know what this code does (because the variables aren't descriptive)

client/examples/repl/command.rs
semenov-vladyslav: You mean deprecated function? It might be used somewhere else. Not sure if I'm authorized to delete something extra.

client/src/procedures/primitives.rs
```diff
@@ -484,20 +490,41 @@ impl DeriveSecret<1> for Slip10Derive {
...
+        let (extended_bytes, chain_code) = match self.input {
             Slip10DeriveInput::Key(_) => {
                 let r = &*guards[0].borrow();
-                let ext_bytes: &[u8; 64] = r
+                let ext_bytes: &[u8; 65] = r
                     .try_into()
                     .map_err(|_| FatalProcedureError::from("bad slip10 extended secret key size".to_owned()))?;
-                slip10::ExtendedSecretKey::try_from_extended_bytes(self.curve, ext_bytes)
-                    .and_then(|parent| parent.derive(&self.chain))
+                match self.curve {
```
DaughterOfMars, line 499: Would it be too much to add a generic key type to the `Slip10Derive` rather than have `Curve`?

semenov-vladyslav: Yes, actually. First, `hazmat` traits are inaccessible because `hazmat` module is private. Second, it's common for stronghold to have runtime types (see [`KeyType`](https://github.com/iotaledger/stronghold.rs/blob/snapshot-migration-v3age-2.0/client/src/procedures/primitives.rs#L325) and [`AeadCipher`](https://github.com/iotaledger/stronghold.rs/blob/snapshot-migration-v3age-2.0/client/src/procedures/primitives.rs#L319) enums, [`GenerateKey`](https://github.com/iotaledger/stronghold.rs/blob/snapshot-migration-v3age-2.0/client/src/procedures/primitives.rs#L587) and [`PublicKey`](https://github.com/iotaledger/stronghold.rs/blob/snapshot-migration-v3age-2.0/client/src/procedures/primitives.rs#L612) procedures). We can make `hazmat` module public in iota-crypto-0.20.1, but it would be difficult to completely fix stronghold.

client/examples/repl/command.rs
semenov-vladyslav: Honestly, 90% of stronghold deserves to be deleted.

client/src/security/keyprovider.rs
```diff
@@ -77,25 +77,18 @@ impl KeyProvider {
     ///
     /// assert_eq!(key, expected, "Key does not match expected {:?}", key);
     /// ```
+    #[deprecated = "this constructor promotes misuse of long passphrases"]
```
thibault-martinez: We're doing a major bump, I'd say we're allowed to just remove this

client/src/security/keyprovider.rs
```diff
@@ -258,6 +244,7 @@ impl KeyProvider {
     ///
     /// assert_eq!(key, expected);
     /// ```
+    #[deprecated = "this constructor promotes misuse of salt in KDF"]
```
thibault-martinez: Same

semenov-vladyslav: Removed

client/src/security/keyprovider.rs
semenov-vladyslav: Removed

.changes/snapshot-migraion-v3age-zeroize.md
```diff
@@ -0,0 +1,13 @@
+---
+
+"iota-stronghold": minor
+"stronghold-engine" : minor
+"runtime" : minor
```
thibault-martinez: I think the crate is called `stronghold-runtime`

client/src/security/keyprovider.rs
```diff
@@ -40,64 +40,6 @@ impl TryFrom<Vec<u8>> for KeyProvider {
...
-    /// assert!(buffer.is_ok(), "Unlocking the keyprovider failed");
-    ///
-    /// // it is safe to unwrap buffer
-    /// let buffer = buffer.unwrap();
-    /// let buffer_ref = buffer.borrow();
-    /// let key = buffer_ref.deref();
-    ///
-    /// assert_eq!(key, expected, "Key does not match expected {:?}", key);
-    /// ```
-    pub fn with_passphrase_truncated<P>(mut passphrase: P) -> Result<Self, ClientError>
```
thibault-martinez, line 80: This is .change worthy

client/src/security/keyprovider.rs
```diff
@@ -214,65 +149,6 @@ impl KeyProvider {
...
-    ///
-    /// // borrow the inner key
-    /// let buffer_ref = buffer.borrow();
-    ///
-    /// // deref the key
-    /// let key = buffer_ref.deref();
-    ///
-    /// assert_eq!(key, expected);
-    /// ```
-    pub fn with_passphrase_hashed_argon2<P>(mut passphrase: P, mut salt: P) -> Result<Self, ClientError>
```
thibault-martinez, line 261: same

.changes/snapshot-migraion-v3age-zeroize.md
semenov-vladyslav: Done

client/src/security/keyprovider.rs
semenov-vladyslav: Done

client/src/security/keyprovider.rs
semenov-vladyslav: Done

.changes/snapshot-migraion-v3age-zeroize.md
DaughterOfMars, line 1: typo in file name

client/examples/repl/command.rs
semenov-vladyslav: Done

.changes/snapshot-migraion-v3age-zeroize.md
semenov-vladyslav: Never noticed it. Fixed.

engine/src/snapshot/migration/v3.rs
```diff
@@ -0,0 +1,51 @@
...
+    let work_factor = age::WorkFactor::new(age::RECOMMENDED_MINIMUM_ENCRYPT_WORK_FACTOR);
+    let age = age::encrypt_vec(password, work_factor, plain).map_err(|_| Error::RngFailed)?;
+    output.write_all(&age[..])?;
+    Ok(())
+}
+
+/// Atomically encrypt, add magic and version bytes as file-header, and [`write`][self::write] the specified
+/// plaintext to the specified path.
+///
+/// This is achieved by creating a temporary file in the same directory as the specified path (same
```
thibault-martinez: I don't see any temporary being created?

.changes/snapshot-migration-v3age-zeroize.md
```diff
@@ -0,0 +1,13 @@
+---
+
+"iota-stronghold": minor
+"stronghold-engine" : minor
+"stronghold-runtime" : minor
+"vault" : minor
+"snapshot" : minor
```
thibault-martinez: The last 2 are not managed by covector, see the list here https://github.com/iotaledger/stronghold.rs/blob/dev/.changes/config.json#L73

engine/src/snapshot/migration/v3.rs
semenov-vladyslav: Removed old comment. We don't want any temp files during migration.

.changes/snapshot-migration-v3age-zeroize.md
semenov-vladyslav: Right. Removed.

engine/runtime/src/memories/noncontiguous_memory.rs
DaughterOfMars: So we're just not gonna change this?

engine/runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -90,14 +90,14 @@ impl LockedMemory for NonContiguousMemory {
...
+        let h = &blake2b::Blake2b256::digest(r);
+        xor_mut(&mut m, h, NC_DATA_SIZE);
```
thibault-martinez: I guess it's fine but just to be sure, the params are not in the same order as before, is that fine?

engine/runtime/src/memories/noncontiguous_memory.rs
```diff
@@ -108,8 +108,8 @@ impl NonContiguousMemory {
             return Err(NCSizeNotAllowed);
         };
         let random = random_vec(NC_DATA_SIZE);
-        let digest = blake2b::Blake2b256::digest(&random);
-        let digest = xor(&digest, payload, NC_DATA_SIZE);
+        let mut digest = blake2b::Blake2b256::digest(&random);
+        xor_mut(&mut digest, payload, NC_DATA_SIZE);
```
thibault-martinez: Because now they are in the same order

engine/runtime/src/memories/noncontiguous_memory.rs
semenov-vladyslav: r - randomness; hr=h(r) - hash of randomness; m - secret message.
shards = (r, m^hr)
To open message from shards: m = shard_2 ^ h(shard_1) = m^hr ^ h(r) = m^hr^hr

engine/runtime/src/memories/noncontiguous_memory.rs
semenov-vladyslav: `xor_mut` does not allocate extra memory. Here, digest = hr; after xor_mut digest = m^hr = shard_2.

thibault-martinez (approved):

engine/runtime/src/memories/noncontiguous_memory.rs
DaughterOfMars: It's almost like these variables should be named appropriately :trollface:

DaughterOfMars (approved):
Much better, thank you.

PR #475, opened 2023-05-26
Slip10Derive extended bytes fix

semenov-vladyslav:
# Description of change

Store/accept 64-byte SLIP10 extended secret bytes. The leading byte (of a full 65-byte extended bytes) is zero, so we skip it. The next 32 bytes are secret keys. This allows to use (64-byte) extended secret bytes as a secret key directly in other operations such as PublicKey, Sign, etc..

Be sure to reference any related issues by adding `fixes issue #`.

## Type of change

- [ ] Documentation Fix

Stronghold doesn't have a mechanism to run a fixed-value tests. `usecase_secp256k1_slip10_derive_key` detects inconsistencies with slip10 key derivation indirectly via extended public key derivation using stronghold and crypto and comparing the two results. It may not catch all issues with extended secret key derivation.

## Change checklist

- [ ] I have followed the contribution guidelines for this project
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/.changes/secp256k1.md
+++ b/.changes/secp256k1.md
@@ -5,4 +5,4 @@
 ---
 
 Secp256k1 ECDSA + SLIP-10 support added.
-Bump `iota-crypto` version to 0.20.0.
+Bump `iota-crypto` version to 0.20.1.
--- a/bindings/native/Cargo.toml
+++ b/bindings/native/Cargo.toml
@@ -23,7 +23,7 @@ iota_stronghold         = { package = "iota_stronghold",   path = "../../client/
 engine                  = { package = "stronghold_engine",  path = "../../engine", version = "1.0.0" }
 tokio                   = { version = "1.15.0", features = ["full"] }
 base64                  = { version = "0.13.0" }
-iota-crypto = { version = "0.20.0", default-features = false, features = [
+iota-crypto = { version = "0.20.1", default-features = false, features = [
   "aes-gcm",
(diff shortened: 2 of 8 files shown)


engine/src/snapshot/logic.rs
```diff
@@ -87,7 +87,7 @@ pub enum WriteError {
 pub fn encrypt_content<O: Write>(plain: &[u8], output: &mut O, key: &Key) -> Result<(), WriteError> {
     let work_factor = age::RECOMMENDED_MINIMUM_ENCRYPT_WORK_FACTOR;
     // `work_factor = 1` is intentionally just for development.
-    // let work_factor = 1;
+    let work_factor = 1;
```
Thoralf-M: Undo

Thoralf-M (changes requested):
missing changes file

engine/src/snapshot/logic.rs
semenov-vladyslav: Yeah, always forget about it. Undid commit.

semenov-vladyslav:
Do we need a changes file? 2.0 hasn't been released yet, and it's a bug fix, not a breaking change actually.

Thoralf-M:
Ah no you're right, no changes file needed, I thought there was a release already

Thoralf-M (approved):

thibault-martinez (approved):

PR #497, opened 2023-10-06
Fix compilation for armv7-unknown-linux-gnueabihf target

Thoralf-M:
# Description of change

Fix compilation for armv7-unknown-linux-gnueabihf target.
`nix::unistd::sysconf(nix::unistd::SysconfVar::PAGE_SIZE)` returns i32 instead of i64 there, that's why the conversion is needed.
Also applied two Clippy suggestions.

https://github.com/iotaledger/iota-sdk/issues/1161

## Type of change

`cargo check --target armv7-unknown-linux-gnueabihf`

### DIFF
--- a/engine/runtime/src/memories/frag.rs
+++ b/engine/runtime/src/memories/frag.rs
@@ -25,6 +25,7 @@ use zeroize::Zeroize;
 // This is the page size for most linux systems
 pub static FRAG_MIN_DISTANCE: usize = 0x1000;
 const MAX_RETRY_ATTEMPTS: usize = 10;
+const DEFAULT_MEMORY_PAGE_SIZE: nix::libc::c_long = 0x1000;
 
 /// Fragmenting strategy to allocate memory at random addresses.
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


engine/runtime/src/memories/frag.rs
```diff
@@ -218,8 +218,9 @@ where
 
         let default_page_size = 0x1000i64;
 
+        #[allow(clippy::useless_conversion)]
         let pagesize = nix::unistd::sysconf(nix::unistd::SysconfVar::PAGE_SIZE)
-            .unwrap_or(Some(default_page_size))
+            .unwrap_or(Some(default_page_size.try_into().unwrap()))
```
thibault-martinez: Are these unwraps really safe ? If yes please put a comment

Thoralf-M: Done

semenov-vladyslav: `sysconf` returns `c_long`. Can't we just define `default_page_size` to be this type? Eg. `const DEFAULT_PAGE_SIZE: nix::libc::c_long = 0x1000;`? No need for a conversion, and the compiler will complain if the literal value doesn't fit `c_long` type.

semenov-vladyslav (changes requested):

engine/runtime/src/memories/frag.rs
Thoralf-M: Thanks for the suggestion, changed

semenov-vladyslav (approved):

